PR #490, opened 2026-06-19
fix(engine): libpod spec/response correctness

### LINKED ISSUE #477: fix(engine): libpod spec/response correctness — stop_signal, ports, secrets, stats
Audit found multiple libpod request/response shape defects, verified live on Podman 5.4.2.

- **CRITICAL — `stop_signal` sent as string → HTTP 500.** `SpecGenerator.stop_signal: Option<String>` (`internal/libpod/types/container/spec/mod.rs:42`) but the libpod field is `syscall.Signal` (int). Any service with `stop_signal:` fails to create. Reproduced live: string → `cannot unmarshal string ... into ... syscall.Signal`, integer `15` → 200. Fix: `Option<i64>` + map signal names→numbers in `engine/container/mod.rs:196`.
- **HIGH — `ContainerPort` drops `protocol`/`range`.** `internal/libpod/types/container/response.rs:47` declares only host_ip/host_port/container_port; live API returns 5 fields. Ranged mappings shown as single port. Add `protocol: Option<String>`, `range: Option<u16>`.
- **HIGH — secrets non-atomic delete-then-create.** `engine/secrets/mod.rs:116` does blind DELETE then create; comment stale on 5.4.2 (`replace=true` works on missing). TOCTOU + unguarded delete. Use single `secrets/create?replace=true`.
- **MED — stats `network: null` deserialize fails** + use server-side `containers=` filter. `engine/stats.rs` — `ContainerStat.network` HashMap is null in some frames → `invalid type: null, expected a map`. Add `#[serde(default)]`; pass `containers=` query.

Each verified against Podman 5.4.2.

Jaro-c:
Closes #477

Four correctness fixes for libpod request/response handling, each verified live on Podman 5.4.2.

- **CRITICAL — `stop_signal` sent as a string (HTTP 500).** libpod's `SpecGenerator.stop_signal` is a numeric `syscall.Signal`; a string value returns 500. The field is now `Option<i64>` (`internal/libpod/types/container/spec/mod.rs`), and the compose `stop_signal:` is resolved to its integer number via a new `resolve_stop_signal` helper (`internal/engine/container/resolve.rs`, wired in `internal/engine/container/mod.rs`). Accepts named, numeric, and `SIG`-prefixed values, case-insensitive; an unknown name now returns a clear `ComposeError::Unsupported` instead of being silently dropped. Unit tests cover named/numeric/`SIG`-prefixed/invalid.
- **HIGH — `ContainerPort` dropped fields.** Added `protocol: Option<String>` and `range: Option<u16>` (serde defaults) to `internal/libpod/types/container/response.rs`. `protocol` is now surfaced in the `ps` port column (`host:hport->cport/proto`, `internal/engine/query/mod.rs`).
- **HIGH — non-atomic secret create.** Replaced the blind DELETE-then-create with a single `POST .../secrets/create?name=<name>&replace=true` (`internal/engine/secrets/mod.rs`). `replace=true` overwrites an existing secret and also succeeds when none exists on 5.4.2, so the prior delete and its stale workaround comment are gone. The owned-label guard for `down` deletion is unchanged.
- **MED — stats `network: null` deserialize failure.** libpod sends `"Network": null` for a container with no interfaces, which failed with `invalid type: null, expected a map`. The field now tolerates null/absent → empty map (`internal/engine/stats.rs`). The stats request is additionally scoped server-side via a `containers=<comma-separated names>` query so the daemon samples only the project's containers. Unit tests cover the null/missing-network deserialize and the query builder.

## Verification (local)
- `cargo fmt --all` — pass
- `cargo clippy --all-targets --all-features -- -D warnings` — pass
- `cargo build --all-features` — pass
- `cargo test --lib --all-features` — pass (657 passed, 0 failed)

Podman integration tests were intentionally not run (shared host).

### DIFF
--- a/internal/engine/container/mod.rs
+++ b/internal/engine/container/mod.rs
@@ -11,7 +11,7 @@ use crate::libpod::API_PREFIX;
 use crate::{ports, size};
 
 mod resolve;
-use resolve::{build_env, resolve_links, resolve_volume_name};
+use resolve::{build_env, resolve_links, resolve_stop_signal, resolve_volume_name};
 pub(crate) use resolve::{config_hash, resolve_bind_source};
 
 use super::container_config::{
@@ -183,6 +183,12 @@ impl Engine {
 			tracing::warn!("{warning}");
 		}
 
+		let stop_signal = service
+			.stop_signal
+			.as_deref()
+			.map(resolve_stop_signal)
+			.transpose()?;
+
 		let spec = SpecGenerator {
 			name: container_name.to_string(),
 			image: image.to_string(),
@@ -193,7 +199,7 @@ impl Engine {
 			stdin: service.stdin_open,
 			user: service.user.clone(),
 			work_dir: service.working_dir.clone(),
-			stop_signal: service.stop_signal.clone(),
+			stop_signal,
 			stop_timeout,
 			hostname: service.hostname.clone(),
 			domainname: service.domainname.clone(),
--- a/internal/engine/container/resolve.rs
+++ b/internal/engine/container/resolve.rs
@@ -187,10 +187,88 @@ pub(super) fn build_env(service: &Service, base_dir: &Path) -> Result<Vec<String
 	))
 }
 
+/// Resolve a compose `stop_signal:` value to its numeric `syscall.Signal`.
+///
+/// libpod's `SpecGenerator.stop_signal` is an integer; sending the signal name
+/// as a string returns HTTP 500. Accepts a bare number (`"15"`), a signal name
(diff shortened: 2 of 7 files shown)


Jaro-c:
Dropped the secrets atomic-`replace=true` change: it regresses on Podman builds that reject `replace=true` when the secret does not yet exist (the exact failure the original code comment documents — surfaced as 6 secret/config materialization tests failing in the coverage job). Reverted to the version-robust delete-then-create path. The three substantive fixes remain: integer `stop_signal`, `protocol`/`range` on the port response, and null-tolerant stats network map.

Issue #504, opened 2026-06-19
ci: run the coverage/integration suite against the latest Podman 5.x


Jaro-c:
CI currently exercises the integration suite against the runner-default Podman **4.9.3** (ubuntu-latest), not the latest 5.x — contradicting the "track latest Podman / requires ≥5.0" stance. The reusable `Glyndor/.github` rust-ci just starts the host `podman.socket`.

Findings (validated locally):
- No official static Podman SERVER binary exists (only `podman-remote-static` client); building from source is heavy.
- The official project image `quay.io/podman/stable` = Podman 5.8.2 and runs nested containers under `--privileged`. Running the coverage job INSIDE it gives a single 5.8.2 for both the engine socket and the tests' bare `podman` CLI calls (unified storage — verified). 43/45 representative tests already pass on 5.8.2.
- Blocker: healthchecks need systemd; a job-container has no systemd → resolved by #503 (drive healthchecks on-demand). The `cli_down_rmi_all` test also needs to `podman pull alpine:latest` before tagging (clean-env precondition).

**Scope:**
1. Modify the reusable rust-ci `coverage` job to run in `quay.io/podman/stable` (privileged) when `inputs.podman` (gated, backward-compatible — only podup uses it). Install rust/cargo-llvm-cov/python in-container; start `podman system service`; set `PODMAN_SOCKET`.
2. Bump podup's rust-ci reusable pin to the new `.github` release.
3. Re-apply the atomic `secrets/create?replace=true` (reverted in #490 because CI 4.9.3 rejected it; valid on 5.x).
4. `cli_down_rmi_all`: pull alpine before tagging.

Depends on #503 (health) landing first.

Jaro-c:
**BLOCKED — systemd/cgroup wall (validated in real CI, PR #506).**

The sidecar approach works to START Podman 5.8.2, but the full suite can't pass against a container-hosted Podman:
- **cgroups**: pause/unpause/stats/top require a systemd cgroup manager. A container/nested Podman has no systemd PID1 → `cgroupfs` manager without delegation → `this container does not have a cgroup`. Neither `--privileged` nor `--cgroupns=host`+cgroup-mount fixes it.
- **host-path binds**: file: secrets/configs and `-v` volumes write temp paths on the host; a sidecar server can't resolve them across the fs boundary (`statfs /tmp/... no such file`). Shape A (whole job in the image) fixes this but not the cgroup issue.

[...]

PR #506, opened 2026-06-19
ci: validate latest-Podman reusable (do not merge)

Jaro-c:
Validation-only: points the rust reusable at the Glyndor/.github branch that runs the integration suite against the latest Podman (quay.io/podman/stable, 5.8.x) instead of the runner's 4.9.3. Watching the Coverage job. Will be closed once green; the real change re-pins to a tagged .github release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ permissions:
 
 jobs:
   rust:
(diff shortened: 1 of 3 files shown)


Jaro-c:
Validation done. Sidecar Podman 5.8.2 starts fine, but the full suite needs systemd: pause/unpause/stats/top require a systemd cgroup manager (nested/container podman has none → 'container does not have a cgroup'), and host-path binds don't resolve across the host/sidecar boundary. Same systemd dependency as healthchecks. Documenting on #504.

PR #537, opened 2026-06-22
Ignore local tooling scratch directories

Jaro-c:
Some local editor and tooling scratch directories were ignored only by my machine-global gitignore, so a clone or CI without it could accidentally commit them. Ignoring them in-repo makes the protection portable across machines.


Jaro-c:
These paths are already covered by my global git excludes, so duplicating them in-repo isn't the approach I want to take. Closing.

PR #681, opened 2026-06-27
docs(readme): add the podup brand banner as the header

Jaro-c:
I led the README with the podup brand banner — the same 1280×640 card used for the GitHub social preview — so the project opens with its visual identity. The demo GIF stays below as the functional showcase.

The image is committed to `docs/assets/podup-banner.png` (a repo-relative asset, so it renders on GitHub and crates.io alike); the HTML source stays in the internal brand system.


Jaro-c:
Wrong base — the banner belongs on the active docs/readme-redesign branch, not a separate one. Moving it there.

Issue #1072, opened 2026-07-19
fix: up() leaks the DependencyNotReady wrapper, changing the public error contract


Jaro-c:
Found while classifying which lane failures are real and which are nested-virt noise (#1039). This one is real, and I reproduced it on my own machine — real Podman 5.4.2, no VM, fails in 1.2s:

```
test health_targeting::wait_healthy_times_out ... FAILED
expected HealthCheckTimeout, got: health check timeout for container 't2580245-wht-db-1'
```

The timeout itself works. What changed is the error the library hands back. #1067 introduced `ComposeError::DependencyNotReady(Arc<ComposeError>)` (`internal/error.rs:130`) and wraps the readiness wait in it (`internal/engine/lifecycle/mod.rs:375`). Its `Display` is transparent (`error.rs:287` prints the inner error), so the message is unchanged and nothing looked wrong — but the variant a caller matches on did change:

```rust
matches!(err, ComposeError::HealthCheckTimeout(_))  // true before 1.11.1, false after
```

The CLI is fine: `internal/main.rs:73` peels the wrapper with `innermost()` before mapping exit codes, which is why no user-visible behaviour changed and why this got through review. The break is on the **library** surface, and podup is 1.0.0 with a frozen public API — this shipped in a minor.

Blast radius today is nothing: helmly-agent still pins `tag = "v0.9.0"` and never matches on `ComposeError`. It matters for the contract and for the next consumer, including us when that pin finally moves.

My suggestion, though the call is yours: keep `DependencyNotReady` as the internal mechanism it was designed to be, and peel it at the library boundary so `up()` returns the same variants it always did. Changing the test to accept the wrapper would be the other option, but that ratifies a breaking change nobody intended to make.

**The part that worries me more than the bug.** This test has been failing in every lane run on **both** majors since #1067 merged on 18 July. Podman 5 went from 145 passed / 10 failed to 144 / 11 with that merge and has stayed there. Nothing went red, because the lane's only hard gate is a pass-count floor of 115 and 144 clears it easily. A real regression was absorbed as nested-virt noise for a day and a shipped release, exactly as #1039 predicts. The identity data is there — the gate just does not look at it.

Jaro-c:
Fixed in #1073, landed on develop.

[...]

Issue #1039, opened 2026-07-17
reduce podman-lane nested-virt flakiness so the gate can tighten beyond a floor


Jaro-c:
The live lane currently gates on a single signal — the core pass count must meet `.github/podman-baseline-tests`. That catches a regression that breaks previously-passing tests, but it can't catch a regression confined to the archive/streaming/lifecycle subset, because the nested-virt environment already fails a broad, variable set of those tests for reasons that aren't podup bugs.

I tried and reverted a stricter gate (fail when failures exceed the connection-drop flakes). Measured on the lane, it's not viable as-is:

- Podman 5: the same ~10 streaming/attach/run tests fail every run, with **zero** connection-drop signatures. Stable, but not signature-detectable.
- Podman 6: **9 to 34** failures run to run, spanning core lifecycle commands (stop/start/up/down/exec) depending on runner load.

So neither the connection-drop heuristic nor a per-test allowlist separates a real regression from the noise without allowlisting the core commands — which would defeat the gate.

The prerequisite is reducing the flakiness itself so the failure set is small and stable, then a per-test allowlist becomes safe. Candidate directions:
- Give the VM more headroom (memory/CPU/IO) or serialize the suite (`--test-threads=1`) so streaming tests don't drop connections under load.
- Retry at the per-test level (nextest retries) rather than re-running the whole suite once, so a transient doesn't consume the whole budget.
- Root-cause the Podman-6 variance specifically — why it ranges 9–34 while Podman 5 is stable at 10.

Once the set is small and stable, gate on the failing-test *identity* against a committed known-flaky allowlist: any failure outside it is a real regression.

Jaro-c:
Data from validating 6.0.1 (#1034) that changes the premise here, so I am parking it where the work is.

I pulled the failing-test identities out of the last five Podman 6 lane runs (the identities have been logged since #1065). Eight tests fail in **every one of them**:

```
commands_networking::engine_cp_to_container_uploads_file
cp_flags::engine_cp_follow_link_uploads_target_contents
cp_flags::engine_cp_to_container_renames_a_single_file
health_targeting::wait_healthy_times_out
run_flags::engine_run_applies_volume_publish_and_interactive
watch_tests::watch_restart_container
watch_tests::watch_sync_creates_missing_target_directory
watch_tests::watch_sync_file_to_container
```

[...]

PR #1074, opened 2026-07-19
ci(podman-lane): gate on which tests fail, not how many

Jaro-c:
Toward #1039 — this is the gate half, not the flakiness half.

The count gate could not see the regression it existed to catch. #1072 broke one test and left the lane at 144 passing against a floor of 115, so nothing went red and the bug shipped in 1.11.1. The identities have been logged since #1065; the gate simply never read them.

The lane now fails when a failing test is **not** on that major's classified list, and the list is derived from the major the VM actually booted — so gating a new major is dropping in a `podman-known-failures-<major>` file, with no edit to the workflow.

**Podman 5 gets a list.** Every one of its ten entries was classified by running the same suite against real Podman on a physical host: **155 passed, 0 failed**, while the lane reports 10 failures for the same code. Those ten are the VM. The file records that reasoning, and says plainly that adding a line is a claim requiring a run rather than a guess.

**Podman 6 deliberately does not get one.** Its failing set carries a variable tail — identities seen in 1-of-5 runs — so a list built from observation would redden innocent pull requests, which is exactly what forced the floor from 120 down to 115. I would rather gate honestly on one major than pretend to gate on two. The gap is written into both the file and the workflow so nobody "finishes" it by guessing; classifying Podman 6 stays on #1039.

The floor stays for both majors as a second, cruder signal: it catches a suite that only half-ran, which comparing identities cannot.

## Verified before claiming

The last change to this gate shipped with a commit message saying it failed closed when it did not, so I ran the logic against real data first:

| case | expected | result |
|---|---|---|
| today's real Podman 5 failing set | pass | pass, 10 entries matched |
| the same set plus #1072's test | **fail** | fails, names the unexpected test |
| no failures at all | pass + stale notice | passes, lists all 10 as stale |
| Podman 6 (no list) | floor only | falls through |

The second row is the point: the regression that shipped would now turn this lane red on its own pull request.

`shellcheck` clean on the step body.

### DIFF
--- a/.github/podman-known-failures-5
+++ b/.github/podman-known-failures-5
@@ -1,0 +1,34 @@
+# Integration tests that fail on the Podman 5 nested-virt lane for environmental
+# reasons, not because podup is broken. The lane fails when a failing test is NOT
+# on this list — a new identity is a real regression, whatever the pass count says.
+#
+# Every entry below was classified by running the SAME suite against real Podman
+# on a physical host (5.4.2, no VM, serial): 155 passed, 0 failed. The lane
+# reports 10 failures for the same code, so these ten are the environment.
+#
+# Adding a line here is a claim that a test cannot pass under nested-virt. Make
+# that claim with a run, not a guess — the whole point of this file is that the
+# 2026-07 regression (#1072) hid inside a count-only gate for a day and a
+# release. Remove a line the moment its test starts passing in the lane.
+#
+# Podman 6 is deliberately NOT gated this way, and this is not an oversight to
+# fix: its failing set carries a variable tail (identities that appear in 1-of-5
+# runs), so a list built from observation would redden innocent pull requests —
+# the failure mode that already forced the pass-count floor down from 120 to 115.
+# Podman 6 keeps the floor until its set is both classified and stable (#1039).
+
+# The `run` family: needs a controlling terminal the VM's serial console cannot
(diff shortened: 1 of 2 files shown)


Jaro-c:
The gate false-reddened on its own first run — worth writing down, because the cause was not the classified list.

It reported `run_flags::engine_run_starts_dependencies_by_default` as unexpected. That test is in the file, and always was. The tell is that it is the **last** element of the comma-separated identity list: the VM reports over `-serial stdio`, so every line of its log arrives CRLF-terminated, and the final identity keeps a trailing `\r` that matches nothing.

The counts never showed this because they are extracted with numeric patterns, which drop the CR silently. The identity list is the first thing in this lane to compare strings.

I got the diagnosis wrong first: I assumed locale collation between `sort` and `comm`, patched that, and then could not reproduce it under either `C` or `en_US.UTF-8`. Feeding the same list with a trailing CR reproduces the exact test CI named, and stripping it clears it. The `LC_ALL=C` pin stays as defensive hygiene, with a comment saying plainly that it was not the cause.

Also fixed the two `shellcheck` SC2001 style hits that reddened `workflow-lint` (`sed 's/^/  /'` → `awk`).

The irony is not lost: a gate built to stop failures being explained away nearly got explained away by its own author on day one. It stays red until the run is green.

Issue #1097, opened 2026-07-20
fix: cp into a container fails on Podman 6


Jaro-c:
`podup cp` into a container, and `podup watch` sync with it, do not work on Podman 6. We declare support for Podman 6, and Fedora, Arch and Manjaro ship it today.

## What I measured

A Fedora rawhide VM on real hardware (KVM, not nested), Podman **6.0.1** — the same version the lane's Podman 6 leg runs, without the nested virtualisation:

```
test result: FAILED. 147 passed; 8 failed
```

Five of the eight are one family, one signature:

```
commands_networking::engine_cp_to_container_uploads_file
cp_flags::engine_cp_follow_link_uploads_target_contents
cp_flags::engine_cp_to_container_renames_a_single_file
watch_tests::watch_sync_creates_missing_target_directory
watch_tests::watch_sync_file_to_container
```

all failing with `Podman(Hyper(hyper::Error(IncompleteMessage)))` — the connection closing before the response completes. `watch` sync is on the list because it copies through the same path.

The same suite, same commit (1.12.0), on real **Podman 5.4.2**: **155 passed, 0 failed**.

## It is ours, not Podman's, and not the environment

- **`podman cp` itself works** in that same VM: file copied, contents read back.
- **Deterministic**: 3 runs, 3 failures.
- **Direction-specific**: `GET /archive` (container → host) passes — `cli_cp_from_container` and `engine_cp_from_container_extracts_file` are both green. Only the `PUT /archive` upload fails.
- **Not a recent regression**: v1.10.1 fails identically, so this predates the 1.11.x streaming work. podup has evidently never supported this on Podman 6.

## Why nobody noticed

`IncompleteMessage` is exactly the signature `podman-lane` treats as a nested-virt connection drop. On the Podman 6 leg these five have been counted as environmental noise, and the leg gates only on a pass-count floor of 115 out of 155 — so five real failures never came close to turning it red.

That is the concrete cost of a count gate, and it is the second time today it has hidden something: #1072 shipped the same way. The Podman 5 leg now gates on failing-test identity (#1074); Podman 6 does not, because its set was never classified. **This is that classification, and it turns out the set was not noise.**

## Where to look

[...]

PR #1101, opened 2026-07-20
fix: report an autostart uninstall that leaves a unit running, and document watch's exit status

### LINKED ISSUE #1080: fix: several commands exit 0 on a failure path
The 2026-07 sweep (#1044) fixed the lifecycle paths and they hold — `down`, `rm`, `start`, `restart`, `kill`, `scale` and `remove_orphans` all aggregate properly now. These are the ones it did not reach.

**`ps` swallows any parse error.** `internal/main.rs:322` uses `parse_files_with_env_files(...).ok()`. Measured: a malformed `compose.yaml` makes `podup ps` print a bare header and exit **0**, where docker compose exits 1 and every other podup command on the same file exits 1. The repo's own policy comment at `main.rs:404-412` says only a *missing* file should be tolerated for label-only commands. A deploy gate or health check built on `compose ps` reads "nothing running, no error" from a file that does not parse.

**`down --rmi`** — `internal/engine/lifecycle/mod.rs:709` warns and `:712` returns `Ok(())` unconditionally. It is the only non-aggregating arm left in a function whose network and volume loops (`:624`, `:660`) do feed `first_err`.

**`build` extra tags** — `apply_extra_tags` (`internal/engine/build/mod.rs:470-489`) returns `()`; it structurally cannot report. Called without `?` at `:413`. Every `build.tags:` that fails to apply is a log line and `build` exits 0 with the tags missing.

**`logs`** (`internal/engine/query/mod.rs:242`, `:267`, `:315`, plus `Err(_) =&gt; break` at `:293`/`:340`) — once service names validate, nothing can make it non-zero. Against a dead socket it warns per container and exits 0.

**`watch`** (`internal/engine/watch/mod.rs:215`, `:105`, `:147`, `:310`) — every sync/rebuild/restart/exec failure for the whole session is a warning; the loop returns `Ok(())`. Docker compose behaves similarly, so this one is arguable — but it is total.

**`stats`** (`internal/engine/stats.rs:286-294`) — a socket death mid-stream ends the command at `debug!` level, invisible at the default `warn` floor.

**`events`** (`internal/engine/events.rs:53-58`) — malformed frames warn and continue.

**`autostart uninstall`** (`internal/autostart/mod.rs:249`, `internal/autostart/quadlet.rs:210`) — a unit that fails to stop or disable is left running silently; only `remove_file` and `daemon-reload` can fail the command.

**`config --hash`** (`internal/startup/config_render.rs:132`) — `serde_json::to_vec(svc).unwrap_or_default()` yields the SHA-256 of the empty string on a serialisation failure. A stable-looking but wrong hash, exit 0. A deploy pipeline keyed on `config --hash` would quietly stop detecting changes.

**And the broadest one — the panic hook.** `internal/startup/mod.rs:134-137` decides "this was a broken pipe" by substring-matching the *whole* panic message:

```rust
lower.contains("broken pipe") || lower.contains("os error 32")
```

Any panic anywhere whose message happens to contain either string — an `.expect()` on an io::Error that wrapped EPIPE, a Podman error quoting a downstream broken pipe — exits **0** with no output. With `panic = "abort"` there is no unwinding, so this hook is the only thing between a panic and the exit status.

Related but separate, worth deciding at the same time: Ctrl-C on an attached `up` returns 0 where docker returns 130 (`internal/engine/query/inspect.rs:459-475`).

Jaro-c:
Partially closes #1080. **Does not close it** — see the scope note at the end.

## What landed here

**`autostart uninstall` no longer reports success while a unit keeps running.** Both modes discarded the systemctl result with `let _ =`, so a unit that refused to stop was left running while uninstall printed its removal lines and exited 0. The operator believes the stack is gone; it comes back at the next boot.

Idempotence is kept, but by acting and checking the result rather than probing a proxy first:

- Quadlet mode stops unconditionally. Measured: `stop` on a unit whose fragment is on disk exits 0 in every state, and every unit in that loop has its fragment on disk. An `is-active` probe is worse than redundant — a service still coming up reports `activating`, which `is-active` calls a failure, so a stack caught mid-start would be skipped and left running while its units were deleted.
- Service mode disables whenever systemd knows the unit, keyed on `is-active` exit **4** ("no such unit") — a number, not localised prose. The unit *file* is not a usable proxy: measured, a fragment deleted out of band leaves the unit loaded, enabled and running, and `disable --now` is the only thing that clears it.

Verified end to end on real systemd: install, `rm` the fragment out of band, `daemon-reload`, then `podup autostart uninstall` takes it from `active` to `inactive`. `--purge` now runs before the failure surfaces, since by then the unit files are gone and `installed_mode` would no longer recognise the project.

**`stats` failures are visible.** They were logged at `debug!`, invisible at the default `warn` floor. The exit code is unchanged.

**`watch`'s exit status is documented** as the one that means nothing — a failed sync/rebuild/restart/exec is a warning and the session still exits 0. That matches docker compose and is deliberate; what was missing was saying so, in the exit-status table and on the library entry point.

Plus a mechanical split: the autostart tests moved to their own file after the new cases pushed `mod.rs` one line over the 500-line hard limit.

## What I backed out, and why

I tried twice to make `logs`/`stats`/`events` report a failure instead of exiting 0, and the live lane rejected both.

The second rejection taught me I had misread the first. **The lane prints every failing test, not only the unexpected ones.** Ten of the fifteen it listed are already in `.github/podman-known-failures-5` — the `run` family and the streaming reads nested-virt cannot deliver. Only five were mine, and all five are `logs`.

So it was never the mid-stream error handling I narrowed away in the second attempt. It was the two faults I had called unambiguous — a replica lookup that fails, and a stream that will not open. Under nested-virt on Podman 5.8.1 those happen transiently; turning them into failures fails commands that worked. Real Podman 5.4.2 here is green on the same commit, which is the whole reason the lane exists.

A third attempt without new information would be a guess. `logs` is back to `develop`'s behaviour untouched, and the question moves to **#1104**, which is specifically about podup not being able to tell a finished libpod stream from a broken one.

## Scope: #1080 stays open

Of its ten points, five were already fixed by #1088 before I started — I audited each against `develop` HEAD and posted the table on the issue. Of the remaining five, this PR closes `autostart uninstall` and documents `watch`. `logs`, `stats` and `events` exit codes are deferred to #1104.

## Test plan

`cargo test --lib` 1260/1260, clippy `--all-targets --all-features` clean, `RUSTDOCFLAGS="-D warnings" cargo doc` clean. Full integration suite green locally against real Podman 5.4.2.

Four new autostart tests using a scriptable `SystemCtl` double (the existing fake always returns 0, so it cannot express "not running" or "the stop was refused"), including one asserting `is-active` is **not** probed in the quadlet path so the bad gate cannot creep back. **All four verified to fail without the fix** — the quadlet pair by stashing the source, the service-mode pair by reverting just the `uninstall` body, since those tests live inside `mod.rs`.

Part of #1080


Jaro-c:
Reworked the autostart commit — the first version of it was wrong, and wrong in the same way the bug it was fixing was wrong: I gated on a proxy instead of doing the thing and checking whether it actually failed.

**What I got wrong.** I measured `systemctl --user stop` on a unit with *no fragment at all* (exit 5, "not loaded") and generalised from it. But that case cannot occur in `uninstall_quadlet` — the loop iterates `.container` files it just found on disk. Re-measured properly:

```
fragment on disk, never started:
  systemctl --user is-active --quiet <unit>   -> exit 3
  systemctl --user stop <unit>                -> exit 0
```

So the `is-active` gate bought nothing, and it actively broke things: a service still coming up reports `activating`, which `is-active` calls a failure. A stack caught mid-start — boot/login start, a slow image pull, a unit sitting in `activating (auto-restart)` — would have been skipped and left running while uninstall deleted its units and exited 0. That is precisely the contract the commit claimed to establish.

The stop is now unconditional, through `checked`.

**Same mistake in service mode.** Gating `disable --now` on the unit file's presence deletes the only recovery path for "fragment gone, unit still loaded". Measured:

```
unit started, fragment rm'd, daemon-reload:
  is-active                       -> active
  disable --now                   -> exit 0, removes the .wants/ symlink and stops it
  is-active --quiet (unknown unit) -> exit 4
```

So the gate is now `is-active` exit **4** — "no such unit", a number rather than localised prose. Verified end to end against real systemd: install, `rm` the fragment out of band, `daemon-reload`, then `podup autostart uninstall` takes it from `active` to `inactive`. Under the previous version it printed "already removed" and left it running.

Two more, same family:

- A probe that cannot be spawned now counts as "known", so the answer to "I could not ask" is to act and report, not to assume there is nothing to do.
- `autostart uninstall --purge` no longer skips the purge when the uninstall reports a failure. By the time it can fail the unit files are gone and `installed_mode` no longer recognises the project, so short-circuiting skipped the teardown exactly when the stack was still up and most needed it. It purges first, then surfaces the failure.

Tests updated to the corrected contract, including one asserting `is-active` is **not** probed in the quadlet path, so the gate cannot creep back in. 1260/1260 green, clippy clean.

Also rebased onto `develop` so the rendered diff no longer looks like it reverts the #1097 doc notes.

Issue #1104, opened 2026-07-20
fix: podup cannot tell a finished libpod stream from a broken one


Jaro-c:
Split out of #1080, where I hit it and had to back out.

## The problem

Every streaming command (`logs`, `stats`, `events`, `attach`, and `run`, which streams its output) reads a `BoxStream` whose contract looks unambiguous:

```rust
Some(Err(e)) => return Err(PodmanError::from(e)),
None         => return Ok(None),
```

A clean end is `None`; an `Err` is a transport fault. So an `Err` ought to mean the socket died and the command should exit non-zero.

**It does not.** I made those commands report a mid-stream `Err` as a failure and the live lane turned red: fifteen tests on the Podman **5.8.1** leg — six `logs`, six `run`, `attach`, `events` — every command that streams. The identical suite is green on real Podman **5.4.2**, and was green on the lane before the change.

So on at least some Podman versions libpod ends a *finished* stream in a way hyper surfaces as an error rather than a clean EOF. Reporting it fails commands that worked perfectly.

## Why it matters

Until this is resolved, a genuine mid-stream failure is indistinguishable from a normal end, so:

- `stats` losing its socket mid-sample exits **0** with a truncated sample. A monitor scraping it reads partial data as complete.
- `events` on a feed that dies halfway exits 0.
- `logs` cannot report a stream that breaks after the first line.

#1080 closed the two unambiguous cases (a replica lookup that fails, a stream that will not open — which is what a dead socket hits), and deliberately left this one open rather than guessing.

## What to find out

1. **What error is it, exactly?** `IncompleteMessage`, an incomplete chunked body, a reset, something else — and does it differ between 5.4.2, 5.8.x and 6.x? The lane's Podman 6 leg is the other data point.
2. **Is it distinguishable at all?** If a finished stream always ends with one specific hyper error and a dead socket with another, the fix is a predicate. If they are the same error, it is not solvable at this layer and the honest answer is that these commands cannot report mid-stream failure — which should then be documented rather than left implicit.
3. **Does the framing carry the answer?** A multiplexed log stream has 8-byte frame headers; a truncated final frame is detectable in a way a transport error is not.

## Note on the lane

This is also a reminder that the local suite is not the gate. Podman 5.4.2 here is green on the exact commit the lane failed on 5.8.1 — the version difference is the whole bug, and only the nested-virt lane sees it. Same lesson as #1097, from the other direction.

Jaro-c:
[...]

PR #1135, opened 2026-07-20
fix: logs must not report success when it reached nothing

Jaro-c:
`podup logs` against an unreachable engine printed nothing and exited **0** — indistinguishable from a project that simply has no logs. A health check or deploy gate built on `compose logs` read success from an engine that was not there.

`stats` and `events` already exit 1 for the same condition. `logs` was the odd one out:

| socket dead | before | now |
|---|---|---|
| `stats` | 1 | 1 |
| `events` | 1 | 1 |
| `logs` | **0** | **1** |

### The failure was not where I first looked

With a dead socket `logs` never reaches `get_stream`. It fails earlier, while resolving each service's live replicas, and that arm warned and continued — leaving zero targets and a clean `Ok(())`.

### An existing test stopped my first attempt, and it was right

I made any non-404 resolution error fatal. `logs_skips_a_service_whose_replica_resolution_errors_but_still_targets_the_rest` failed, and its doc-comment explains why that tolerance exists: a transient libpod error on one service should not blank the logs of the others. That was a deliberate decision and my change overreached it.

The rule that satisfies both: **tolerate a partial failure, fail only when there is no partial result to protect.** If any service resolved, or any container streamed, `logs` succeeds exactly as before. Only when nothing resolved *and* something went wrong does it return the error.

### Measured against Podman 5.4.2

```
dead socket, nothing resolves     exit 1
one service up, one not created   exit 0, prints the one that is up
normal                            exit 0
service not in the compose file   exit 1
```

Both sides are pinned by tests, and the new one is falsifiable: removing the guard makes `logs_fails_when_no_service_resolves_at_all` fail.

### This is deliberately not #1104

Nothing here classifies how a stream *ended*. Every case is a request that never opened, which needs no predicate for telling a finished stream from a broken one. The `logs`/`stats`/`events` mid-stream behaviour stays blocked on #1104.

Part of #1080

### DIFF
--- a/.github/podman-known-failures-5
+++ b/.github/podman-known-failures-5
@@ -2,33 +2,19 @@
 # reasons, not because podup is broken. The lane fails when a failing test is NOT
 # on this list — a new identity is a real regression, whatever the pass count says.
 #
-# Every entry below was classified by running the SAME suite against real Podman
-# on a physical host (5.4.2, no VM, serial): 155 passed, 0 failed. The lane
-# reports 10 failures for the same code, so these ten are the environment.
-#
 # Adding a line here is a claim that a test cannot pass under nested-virt. Make
 # that claim with a run, not a guess — the whole point of this file is that the
 # 2026-07 regression (#1072) hid inside a count-only gate for a day and a
 # release. Remove a line the moment its test starts passing in the lane.
 #
-# Podman 6 is deliberately NOT gated this way, and this is not an oversight to
-# fix: its failing set carries a variable tail (identities that appear in 1-of-5
-# runs), so a list built from observation would redden innocent pull requests —
-# the failure mode that already forced the pass-count floor down from 120 to 115.
-# Podman 6 keeps the floor until its set is both classified and stable (#1039).
+# 2026-07-20: eight of the ten entries were removed, and they were never about
+# nested virt. Podman's journald log driver could not open its library in this
+# image, so every logs request returned HTTP 500 and the VM served no logs at
+# all. With the file-based driver configured the lane went from 10 failures to
+# 2 — the six run/attach tests and cli_run_subcommand and cli_logs_tail_limits_output
+# all pass now. The claim "cannot pass under nested virt" was made honestly, with
+# a run, and was still wrong: the physical host that proved 155/155 also had a
+# working journald, so the comparison could not see the difference.
 
-# The `run` family (7): needs a controlling terminal the VM's serial console
-# cannot provide. All seven passed in the real-host run that classified them.
-commands_networking::engine_run_command_succeeds
-commands_networking::engine_run_nonzero_exit_returns_run_exited
-run_flags::engine_run_applies_user_workdir_env_and_entrypoint
-run_flags::engine_run_applies_volume_publish_and_interactive
-run_flags::engine_run_no_deps_skips_dependency_startup
-run_flags::engine_run_starts_dependencies_by_default
-cli_commands::cli_run_subcommand
-
-# Streaming reads that depend on timely delivery through the guest's virtio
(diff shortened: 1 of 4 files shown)


Jaro-c:
The lane rejected this, and the reason is worth more than the change itself.

```
Podman(Api { status: 500, message: "failed to obtain logs for Container
't3873-lga-web-1': unable to open a handle to the library" })
```

**The Podman 5 VM cannot serve logs at all.** Podman's journald driver is the Fedora default and `dlopen()`s libsystemd; the VM installs `podman rust cargo gcc git` and nothing that provides it. Every logs request 500s.

**Which means five integration tests have been passing without verifying anything.** They call `engine.logs(...).await.unwrap()`, `logs` swallowed the 500 and returned `Ok(())`, and the unwrap succeeded with no output ever fetched — `logs_all_services`, `logs_from_named_service`, `logs_scaled_service_all_replicas`, `cli_logs_subcommand`, `logs_with_stderr_output`. This PR did not break them; it made them able to fail.

The corroboration is in the known-failures list. Entry 8 is `cli_flags::cli_logs_tail_limits_output` — the one logs test that asserts on its **output**, so the swallowed 500 could never fool it. It has been failing all along, classified as nested-virt. Entry 9, `niche::cli_attach_streams_output_until_exit`, streams logs too.

That file says a line is a claim that a test *cannot pass under nested virt*, to be made with a run and removed the moment the test passes. Two of its ten entries look like they were never about nested virt — they were about a missing library. That matters for #1039, whose premise is that the lane's failures are environmental noise: some of the noise is a fixable configuration error.

Pushed: `systemd-devel` in the VM's package list.

**Falsifiable prediction.** The five logs tests pass, entry 8 starts passing, possibly entry 9 too, and the Podman 5 leg goes from `fail=15` on this branch to **10 or fewer**. If it does not, the diagnosis is wrong and the 500 has another cause.

Jaro-c:
**The `systemd-devel` fix was wrong.** Recording it, since I predicted it would work.

Predicted: zero `unable to open a handle to the library` errors and Podman 5 back to `fail=10`. [Run 29785338714](https://github.com/Glyndor/podup/actions/runs/29785338714):

```
PODUP_SUMMARY pass=145 fail=15 flaky=0
errors matching "unable to open a handle to the library": 21
```

The package installed — `systemd-devel-0:259.7-1.fc44`, `Installing ... 100%` in the run log — and nothing changed. So the missing-symlink diagnosis was wrong, even though the symptom (journald cannot open its library) is real.

I also made a bad call in between: I reported the library errors as gone after grepping a log file that had failed to download and was empty. A grep over an empty file returns zero and reads like a win. Corrected in the run above, which is the real measurement.

### Second attempt, and it is not a guess

Rather than keep theorising about why the go-systemd binding fails in this image, use the driver that has no library to open:

```toml
[containers]
log_driver = "k8s-file"
```

Supported Podman configuration, and podup reads logs through the API either way — so nothing about what the suite exercises changes, except that it now exercises it at all.

### Gated, because silence is the failure mode here

With logs unavailable, the five logs tests passed *without fetching a byte* — `logs` swallowed the 500, returned success, and their `unwrap()` succeeded on nothing. A config file silently ignored would put them straight back to green-and-meaningless, which is worse than a red run.

So the VM prints its resolved driver and the verify step fails the run if it is not `k8s-file`:

```
grep -q 'LOGDRIVER=k8s-file' "$L" || { echo "::error::log driver is not k8s-file — logs tests would pass without reading anything"; exit 1; }
```

Prediction, again falsifiable: zero library errors, the five logs tests pass, and `cli_flags::cli_logs_tail_limits_output` (known-failures entry 8) starts passing — Podman 5 at **10 or fewer**.

Issue #1180, opened 2026-07-26
test: the secrets/env_file tests assert that nothing errored, not that anything happened


Jaro-c:
The empirical CLI sweep of 2026-06-25 left a list of surfaces it never reached:
`attach`, `watch` sync/rebuild, `push` to a real registry, **`secrets`/`env_file`**,
multi-`-f`, `include`/`extends`, and per-command `--help`.

#1174 was in `secrets`. It sat there for the whole life of the feature and no test
noticed, on a lane that runs Fedora with SELinux enforcing and therefore had every
chance to catch it. The environment was not the gap. The assertion was.

Scanning the integration suite for tests whose body contains no `assert` at all
returns **71 of 160**. That number needs a caveat: an `.unwrap()` on `up()` is a
real assertion — it proves the call did not return `Err` — and for a few tests
that genuinely is the whole point. The problem is the ones that name an effect
they never look at:

| test | claims | checked |
|---|---|---|
| `env_file_loaded` | the variable is loaded | nothing — it noted which source line it covered |
| `file_config_bound` | a `file:` config is mounted | nothing |
| `env_config_materialized` | the value materialises | nothing |
| `inline_secret_materialized` | the secret materialises | `cat` exits 0 (passes on an empty file) |
| `inline_config_materialized` | the config materialises | nothing |
| `secret_long_form_ref` | `target`/`mode`/`uid` are applied | `cat` exits 0; mode and uid never read |
| `config_long_form_ref` | an absolute target is honoured | `cat` exits 0 |
| `invalid_secret_name_rejected` | a bad name is rejected | **nothing** — it declares an ordinary secret called `evils`, discards both results with `let _ =`, and says in its own comment that it cannot test what it is named for |

`invalid_secret_name_rejected` is the one worth singling out: it is not a weak
test, it is a test that documents its own uselessness and was kept anyway.

## What to do

Fix the `secrets`/`env_file` slice first, since that is where a bug has already
been found, and prove each strengthened assertion by mutation: break the input,
watch the test go red, restore it. A new assertion that has never failed is
another line of false confidence.

The remaining surfaces from the sweep list stay open after this — `attach`,
`watch` rebuild, `push` (needs a registry in the lane), multi-`-f`,
`include`/`extends`, per-command `--help` — and so do the weak assertions outside
this slice, which the table above only samples.

Jaro-c:
Progress so far, all on `develop`:

| slice | PR | outcome |
|---|---|---|
[...]

Issue #1174, opened 2026-07-25
fix: a file: secret or config is unreadable inside the container on an SELinux host


Jaro-c:
On a host with SELinux enforcing, a `file:` secret or config is mounted into the
container at the right path with the right content, and the container cannot read
it. `up` reports the container as Started, so nothing signals the failure except
the application failing later.

Inline secrets on the same host work. That is the part that makes this ours
rather than "SELinux is strict": podup already routes `content:`/`environment:`
sources through a native Podman secret, which is labelled correctly, and only
`file:` sources keep a plain read-only bind of the host path.

## Measured

Fedora 44 (Podman 5.8.1) and Fedora Rawhide (Podman 6.0.1), both `getenforce` =
`Enforcing`, rootless, podup 3.0.2.

Same project, same target path, only the secret source differs:

```yaml
secrets:
  filesecret:
    file: ./my_secret.txt
```

```
 Container sel-secret-web-1  Started
web-1 | cat: can't open '/run/secrets/filesecret': Permission denied
```

```yaml
secrets:
  inlinesecret:
    content: inline-secret-content
```

```
 Container sel-inline-web-1  Started
web-1 | inline-secret-content
```

The denial is not podup-specific — plain `podman run` reproduces it, so the
mechanism is just the missing relabel:

```
$ ls -Z my_secret.txt
unconfined_u:object_r:user_home_t:s0 my_secret.txt
$ podman run --rm -v $PWD/my_secret.txt:/run/secrets/filesecret:ro alpine cat /run/secrets/filesecret
cat: can't open '/run/secrets/filesecret': Permission denied
$ podman secret create t1 my_secret.txt && podman run --rm --secret t1 alpine cat /run/secrets/t1
file-secret-content
```

Reproduced from `$HOME` and from `/tmp` (`user_home_t` and `user_tmp_t` alike).

## Where it comes from

`make_bind` in `internal/engine/secrets/mod.rs` builds `{resolved}:{target}:ro`
with no mount options beyond `ro`. Nothing downstream adds a relabel, so the
container gets a bind it cannot read whenever the host enforces SELinux. The same
function serves `configs:`, so `configs: file:` has the identical hole.

This lands on Fedora, RHEL and CentOS Stream — the distributions where Podman is
the default container engine and SELinux is enforcing out of the box.

## Why no test caught it

`file_secret_bound` mounts the secret and then only asserts that `up` and `down`
succeed. It never reads the secret from inside the container, so it passes on an
[...]

PR #1183, opened 2026-07-26
test: assert the effect, not just the absence of an error

Jaro-c:
> Replaces #1181, which GitHub closed on its own when the base branch it was
> stacked on was deleted at merge. Same commit, rebased onto `develop` now that
> #1178 has landed.

Nine tests in the secrets, configs and `env_file` slice started a container,
stopped it, and asserted nothing about the thing they are named for. They passed
whether or not it happened.

`env_file_loaded` is the clearest: it wrote `MYVAR=hello`, brought the service up,
brought it down, and carried a comment saying which source line it covered. Line
coverage, not effect. It would have passed with `env_file` dropped on the floor.

This matters because #1174 — a `file:` secret unreadable on every SELinux host —
lived in this exact slice for the whole life of the feature, on a lane that runs
Fedora with SELinux enforcing and therefore had every chance to catch it. The
environment was never the gap. The assertion was. That slice is also on the
2026-06-25 empirical sweep's own list of surfaces it never reached.

Each test now checks the observable result from inside the container through
`exec`, whose non-zero exit surfaces as `RunExited`, so the exit code *is* the
assertion and no sleep stands in for synchronisation.

| test | was | now |
|---|---|---|
| `env_file_loaded` | nothing | reads `$MYVAR` |
| `file_config_bound` | nothing | compares the content |
| `env_config_materialized` | nothing | compares the content |
| `inline_config_materialized` | nothing | content, and pins the container-root default target that distinguishes a config from a secret |
| `inline_secret_materialized` | `cat` exits 0 | compares the content |
| `secret_long_form_ref` | `cat` exits 0 | content plus the `mode: 0400` and `uid` the compose file asks for |
| `config_long_form_ref` | `cat` exits 0 | content at its absolute target |
| `invalid_secret_name_rejected` | **nothing at all** | requires `up` to reject `../evil` |

`cat` on its own exits 0 on an empty file, which is what the two tests that did
exec were really asserting.

`invalid_secret_name_rejected` deserves its own line. It declared an ordinary
secret called `evils`, discarded both results with `let _ =`, and stated in a
comment that it could not test the traversal its name promises. It was not a weak
test; it documented its own uselessness and was kept.

## Test plan

- All nine pass locally against real Podman 5.4.2.
- **Every new assertion proved by mutation.** Breaking the input turns the test
  red with its own diagnostic, then the input is restored:

  ```
  test health_targeting::env_file_loaded ... FAILED
  test build_resources::file_config_bound ... FAILED
  MYVAR from env_file did not reach the container's environment: Err(RunExited(1))
  config /filecfg did not carry the file's contents: Err(RunExited(1))
  ```

  An assertion that has never failed is another line of false confidence, which
  is the whole point of this PR.
- `cargo fmt --check` and `clippy --all-targets` clean.
- The lane runs both majors on Fedora with SELinux enforcing, which is where the
  strengthened `file_config_bound` has teeth.

## Not covered

The rest of the 2026-06-25 sweep list stays open: `attach`, `watch` rebuild,
`push` (needs a registry in the lane), multi-`-f`, `include`/`extends` and
per-command `--help`. So do the weak assertions outside this slice — a scan for
integration tests with no `assert` in the body returns 71 of 160, and #1180
samples the rest.

Refs #1180

### DIFF
--- a/tests/engine_integration.rs
+++ b/tests/engine_integration.rs
@@ -88,6 +88,8 @@ mod cp_flags;
 mod exec_flags;
 #[path = "engine_integration/health_targeting.rs"]
 mod health_targeting;
+#[path = "engine_integration/include_extends.rs"]
+mod include_extends;
 #[path = "engine_integration/lifecycle.rs"]
 mod lifecycle;
 #[path = "engine_integration/niche.rs"]
--- a/tests/engine_integration/build_resources.rs
+++ b/tests/engine_integration/build_resources.rs
@@ -23,7 +23,26 @@ async fn file_config_bound() {
 	let file = parse_str(&yaml).unwrap();
 
 	engine.up(&file).await.unwrap();
+	// Read the config from inside the container. Asserting only that `up` and
+	// `down` succeed passes just as happily when the config is missing, empty or
+	// unreadable — the mount is not the effect, the content is.
+	let read = engine
+		.exec_with_options(
+			&file,
+			"web",
+			vec![
+				"sh".to_string(),
+				"-c".to_string(),
+				"test \"$(cat /filecfg)\" = key=from-file".to_string(),
+			],
+			podup::ExecOptions::default(),
+		)
+		.await;
 	engine.down(&file).await.unwrap();
+	assert!(
+		read.is_ok(),
+		"config /filecfg did not carry the file's contents: {read:?}"
+	);
 }
 
 #[test]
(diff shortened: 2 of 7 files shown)


Jaro-c:
Extended past the original secrets/configs/`env_file` slice, since the same shape kept turning up:

**`engine_cp_to_container_uploads_file`** copied a file into a container, checked only that `cp` returned `Ok`, and tore the project down before it could have checked anything else. That is blind to precisely what #1097 was — libpod accepted the archive, closed the connection with no response, and podup reported success while nothing had arrived. The reverse direction already asserted the extracted file exists; only the upload was unverified.

**`engine_port_returns_binding`** was named for a return value the API does not have. `Engine::port` is `Result<()>` and writes the binding to stdout, so the test could never assert the binding. Renamed to `engine_port_resolves_a_published_port` for what it does check, and the binding is now asserted where it is observable — a CLI test that publishes `127.0.0.1:18081` and requires it in stdout.

Both proved by mutation, same as the rest:

```
test commands_networking::engine_cp_to_container_uploads_file ... FAILED
test stats_flags::cli_port_prints_the_published_binding ... FAILED
cp reported success but /tmp/testfile.txt is missing or has the wrong contents: Err(RunExited(1))
port did not print the host binding it was asked for: "127.0.0.1:18082\n"
```

The port failure printing the binding it actually read is the useful part — the assertion is reading real output, not a placeholder.

Issue #1207, opened 2026-07-27
rootless-netns resolv.conf mount fails on Podman 6, reproducible with plain podman (upstream)


Jaro-c:
Serialising the Podman 6 lane leg removed a container-creation race (#1039, run
30225458799: 31 failures to zero, the signature gone from the log). That fixed
the lane. It did not answer whether the same race is reachable by a user.

The error:

```
podman API error (HTTP 500): rootless netns: mount resolv.conf to
"/run/user/1000/containers/networks/rootless-netns/run/systemd/resolve/stub-resolv.conf":
no such file or directory
```

The missing path is the mount destination inside Podman's own rootless-netns
tree, and the trigger is concurrent container creation against that shared netns.

## Why this is not obviously a test-only problem

The lane creates containers concurrently because it runs tests in parallel. **A
user running `up` on a multi-service project does the same thing** — podup starts
the services in a dependency level concurrently, by design. If the race is
reachable there, capping the lane's thread count hides a defect users can hit
rather than fixing it, and the fix would belong in podup (serialising netns-creating
work) or upstream in Podman.

## What makes this hard to answer

It did not reproduce on a single-level Fedora Rawhide VM with Podman 6.0.1, under
conditions matching the lane. The race may need the extra slowness of nested
virtualisation to widen the window, which would mean the lane is the only place it
currently shows — and a "cannot reproduce" locally proves little.

## Suggested approach

- Drive `up` on a project with many services and no dependency ordering (so every
  container is created at once) on a Podman 6 host, repeatedly, and watch for the
  signature.
- If it appears: decide between serialising netns-creating work in podup and
  reporting upstream, and check whether Podman 5 is affected too.
- If it does not, after enough attempts to mean something, record that and close —
  with the caveat that absence under one set of conditions is weak evidence, which
  is exactly the trap three earlier hypotheses in #1039 fell into.

Filed so the question is not lost behind a lane that now looks green.

Jaro-c:
Measured this on the real Podman 6 VM. **It does not reproduce** — and one
structural finding says why the worry may have been aimed at the wrong shape.

## What I ran

Fedora 45 (Cloud Prerelease), Podman 6.0.1, rootless, single-level virtualisation.
Preconditions for the failing mount verified first, because a VM that lacks them
cannot produce the error whatever you do to it:

```
/etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
systemd-resolved: active
```

[...]

PR #1221, opened 2026-07-27
ci(podman-lane): run the Podman 6 leg at two threads again

### LINKED ISSUE #1214: podman-lane: decide whether the Podman 6 leg keeps running at one thread
The Podman 6 leg runs the integration suite at `--test-threads=1` (#1205), which
is what removed the `rootless netns: mount resolv.conf … stub-resolv.conf: no
such file or directory` tail and let the leg gate on identity with an empty
classified list (#1206). The diagnosis holds — 172/172 on both legs across the
runs since — but the price is now permanent and it is paid on every pull request:

| leg | threads | suite time (nightly 30265944631) |
|---|---|---|
| Podman 5 | 2 | 16m48s → 18m01s wall |
| Podman 6 | 1 | 1634s suite, 29m38s wall |

The lane was already the long pole. Roughly ten minutes of suite time became
close to thirty, so a PR that touches the engine waits half an hour for its
slowest required check, and the nightly holds a runner for the same.

What I want to decide, with a run rather than an argument:

- **Is 1 the right number, or was it just the cheapest way to test the
  hypothesis?** #1205 chose one thread to falsify a cause, not to tune a lane.
  Two threads with the identity gate live is now a *safe* experiment in a way it
  was not before: if the tail returns, the empty list makes it loud instead of
  hiding it inside a pass count. That was the whole reason the gate was tightened
  first.
- **Does the concurrency have to be capped in the harness at all**, or is the
  right fix on the podup side? That question is #1207 — if the race is reachable
  from `up` on a multi-service project, serialising the lane hides a user-facing
  defect behind a clean CI. Answering #1207 may make this issue moot.
- **Is there a cheaper axis than threads** — the VM's CPU/RAM allocation, or
  splitting the suite across two jobs so wall-clock drops without raising
  per-socket concurrency.

Not urgent, and explicitly not a reason to loosen the gate: a slow lane that
tells the truth beats a fast one that does not. But it should be a decision on
record rather than a side effect of an experiment that stayed.

Jaro-c:
Both legs run at two threads again. The experiment that put Podman 6 at one is
over, and its price is not.

## What this is

One thread was how the netns hypothesis got **tested** (#1039): removing the
concurrency took the leg from 31 failures to zero and the `rootless netns: mount
resolv.conf … stub-resolv.conf: no such file or directory` signature disappeared
with them, while Podman 5 stayed at two as the control. That identified the cause.
It was never a decision about how the lane should run, and it cost the leg roughly
ten minutes of suite time turning into close to thirty — on the slowest thing
standing between a pull request and a merge.

## Why it is safe to try now, and was not then

In #1039 that leg gated on a pass-count floor which had already reported it green
through 31 failures (141 passed against a bar of 115). A returning tail would have
been invisible.

Today **both majors gate on identity with an empty classified list** (#1206,
#1215). One failing test reddens the leg, by name. That is exactly the property
that makes an experiment like this cheap: the lane cannot quietly absorb the
answer.

## Both outcomes are results

- **Green** — the serialisation was overpriced, and every engine PR gets about
  fifteen minutes back.
- **Red** — the window is real at this concurrency, in the one environment that
  has ever shown it. That is evidence #1207 cannot get from my hardware: 72 cycles
  and roughly 1,800 container creations across four shapes, including a teardown
  racing a creation across the netns refcount, produced nothing on metal.

Either way this PR's own lane run is the measurement.

## Not reverted

The job and qemu timeouts stay where the serialised runs put them. Headroom is
free, and a run that dies on its own timeout answers nothing.

Closes #1214

### DIFF
--- a/.github/workflows/podman-lane.yml
+++ b/.github/workflows/podman-lane.yml
@@ -21,8 +21,10 @@ permissions:
 jobs:
   podman-vm:
     runs-on: ubuntu-24.04
-    # Raised for the one-thread experiment: serialising the suite costs wall
-    # clock, and a run that dies on the timeout answers nothing (#1039).
+    # Raised when the Podman 6 leg ran serialised (#1039) and deliberately kept
+    # there now that both legs are back at two threads (#1214): a run that dies on
+    # its own timeout answers nothing, and headroom on a lane that only pays for
+    # itself when it is trustworthy costs nothing.
     timeout-minutes: 75
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


Jaro-c:
Closing this unmerged: the change is wrong, and this PR's own lane run was the
thing that nearly sold it.

That run came back green at two threads — 172 passed, 0 failed, no retry. I did
not merge on it, because it changed the thread count *and* the rawhide image
underneath at the same time (#1222), and because this repository has twice written
down what happens when a claim rests on one observation. So I dispatched three more
runs on this branch, on the **same** image that had just gone green:

| run | Podman 6 | `rootless netns` in the log |
|---|---|---|
| 30288756745 | 140 passed, **32 failed** | 4 |
| 30288763334 | 159 passed, **13 failed** | 6 |
| 30288770662 | 139 passed, **33 failed** | 4 |

Three of four red, same image, same commit, with the exact signature:

```
rootless netns: mount resolv.conf to
"/run/user/1000/containers/networks/rootless-netns/run/systemd/resolve/stub-resolv.conf":
no such file or directory
```

and a second shape of the same failure, `rootless netns: deserialize info: EOF`.

So the green run was luck, at roughly one time in four. Had this merged, every
engine pull request would have gone red about three times out of four, and the
first instinct on a red would have been to doubt the gate that was doing its job.

**The serialisation was not overpriced.** The Podman 6 leg keeps its thread of one,
and its thirty minutes are the price of a lane that tells the truth. #1039's
diagnosis now holds across two different rawhide images.

Issue #1298, opened 2026-08-03
ps and images carry no time or size, and their JSON does not either


Jaro-c:
`podup ps` cannot tell you how long a container has been up, and `podup images` cannot tell you how big an image is. Neither can the JSON, which is what makes this more than a rendering choice.

## Measured

Podman 5.7.0 against `docker-compose` v5.1.3 on the same socket.

**`ps`**

```
$ podup ps
NAME             IMAGE                           STATUS  PORTS
up-…-web-1       docker.io/library/alpine:latest running

$ docker compose ps
NAME         IMAGE          COMMAND            SERVICE  CREATED        STATUS        PORTS
dc-…-web-1   …alpine:latest "sleep infinity"   web      7 seconds ago  Up 6 seconds
```

`podman ps` agrees with docker: `Up 5 seconds`.

**`images`**

```
$ podup images
SERVICE REPOSITORY TAG    IMAGE ID
web     alpine     latest d529dd0c6e55

$ docker compose images
CONTAINER       REPOSITORY  TAG     PLATFORM      IMAGE ID      SIZE     CREATED
dcan-…-web-1    alpine      latest  linux/amd64   d529dd0c6e55  8.71MB   6 weeks ago
```

## The JSON does not carry it either

This is the part that decides the severity:

```
ps --format json      → ExitCode, Health, ID, Image, Name, Project, Publishers, Service, State, Status
images --format json  → ID, Repository, Service, Tag
```

If the field were in the JSON and merely absent from the table, this would be a rendering decision and defensible. It is not there at all, so **a script cannot obtain a container's uptime or an image's size from podup by any route**. That is information missing from the model, not from the renderer.

For `images` in particular, size is the main reason anyone opens that table.

## The data is in libpod, verified

Queried directly on the same socket:

```
containers/json → Created:   "2026-08-02T19:25:34.692577272-05:00"
                  StartedAt: 1785716734           (unix epoch)
                  Command:   ["sleep","infinity"]
                  Status:    ""                   ← empty, see below

images/json     → Size:    8710622                (= docker's 8.71MB)
                  Created: 1781568089             (unix epoch)
```

Note `Status` is empty. That is the same shape as **#590** — *"`ps` STATUS column always empty — reads `c.status` (empty on libpod) instead of `c.state`"*. The fix then took `State`, which is the word `running`. The **uptime is not in either field**: it has to be computed from `StartedAt`, so this is not a matter of reading a different column.

## How it could be improved

**`ps`**
[...]

PR #1306, opened 2026-08-03
feat(units): months, and three components by default for sizes and durations

Jaro-c:
The owner's rule for how a span reads, which I had wrong in #1303.

## The window slides, the width does not

Three components, and which units fill them moves with the magnitude:

| span | reads |
|---|---|
| just started | `5s` |
| a minute and a half | `1m 30s` |
| an hour in | `1h 5m 3s` |
| two months | `2mo 1d` |
| past a year | `1y 1mo 5d` |

Same column width at every scale, always carrying the three units that matter
there. The default was two components; it is three.

## Months are back

I took them out of #1303 on the grounds that 28 to 31 days is ambiguous. That
reasoning was mine, written into the issue body — it was never the owner's
requirement, and the owner wants months.

**A month is exactly 30 days**, the same way a year is exactly 365. Neither
consults a calendar, because a duration is an elapsed span and has none.

**They do not divide evenly and the output says so:** 364 days reads `12mo 4d`
while 365 reads `1y`, because twelve thirty-day months are 360 days. The
alternative is a fractional month of 30.4167 days, which no reader can check by
hand. The seam is preferred over arithmetic nobody can verify, and it has its
own test so nobody "fixes" it later without reading why.

**Weeks stay out.** A month at least has a conventional length; a week has none
in a duration context at all, and `9d` already says what `1w 2d` would.

## A test that was not testing what it claimed

The first version of the sliding-window test used five spans, and every one of
them happened to have consecutive units — no gaps. So it passed unchanged
against a formatter that stops at the first empty unit instead of skipping it.
The mutation proving the zero-skip died by a different test entirely, which is
how it surfaced. Two gap rows added: a year and two days has no whole month in
it, and that is the case that tells the two behaviours apart.

## Test plan

- 12 duration tests, including the owner's table above and the 364/365 seam.
- **4 mutations run, 4 killed**: the month set to 31 days, the month rung
  collapsed onto days, the default put back to two components, and the zero-skip
  turned into a stop.
- Full lib suite: 1478 passed.
- No runtime surface yet — nothing calls the duration formatter, so there is
  nothing to measure against a live Podman. The CREATED columns that will call
  it are the next piece of #1298.

Part of #1298

### DIFF
--- a/internal/units/bytes.rs
+++ b/internal/units/bytes.rs
@@ -123,6 +123,21 @@ impl SizeFormat {
 			..self
 		}
 	}
+
+	/// Same ladder, the default number of components.
+	///
+	/// Three, matching [`DurationFormat::default_parts`]. One default across
+	/// both formatters is the point: a reader who has learned what `1h 5m 3s`
+	/// means should not have to learn a second rule for `1GB 512MB 200kB`.
+	///
+	/// This is the default *component count*, not a floor on the output. A value
+	/// with fewer components than that still renders only the ones it has —
+	/// three is what a caller gets when it does not choose, never padding.
+	///
+	/// [`DurationFormat::default_parts`]: super::DurationFormat::default_parts
+	pub(crate) const fn default_parts(self) -> Self {
+		self.with_parts(DEFAULT_PARTS)
+	}
 }
 
 /// How many decimal places a single-unit value shows.
@@ -166,6 +181,12 @@ impl Precision {
 		}
 	}
 }
+
+/// The default component count for both formatters.
+///
+/// Lives here and is read by the duration side too, so the two cannot drift to
+/// different defaults without the change being visible in one place.
+pub(crate) const DEFAULT_PARTS: usize = 3;
 
 /// Render `bytes` as a human-readable size.
 ///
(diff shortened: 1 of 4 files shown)


Jaro-c:
Extended after the owner confirmed: **minimum three, in time and in size alike.**

The component count now lives in one constant that both formatters read, so they cannot drift to different defaults. `SizeFormat::default_parts()` mirrors `DurationFormat::default_parts()` — `1TB 512GB 200MB` next to `1h 5m 3s`, one rule to learn instead of two.

Three is what a caller gets when it does not choose, **not a floor on the output**: a value with fewer components than that renders only the ones it has, never padded out to length. Pinned by its own test, and by a mutation that adds the padding and dies.

The other reading of "minimum three" — three significant digits on a single-unit size — is already what #1305 does for the `images` SIZE column, measured against podman and docker compose. Both readings are satisfied.

Mutations now 8, all killed: the shared constant set to two, the duration side going back to a literal instead of reading it, the month at 31 days, the month rung collapsed onto days, the default put back to two, the zero-skip turned into a stop, and the short-value padding added.

Issue #1361, opened 2026-08-07
fix(security): tighten path and value sanitization for compose-driven file inputs


Jaro-c:
## What's wrong

Several compose fields let the operator point podup at host-resident files or pass user-controlled values into places downstream consumers re-parse. The threat model accepts this (compose-spec, drop-in fidelity), but the boundary is not always clean: a few values land in places that downstream consumers re-parse, a few paths are unbounded in a way that can grow unbounded, and one Quadlet field was originally thought to be a semantic inversion but is **actually correct per the podman documentation**. Validation against the podman source corrected three of the original findings; the remaining work is hygiene.

## Validation corrections (from the podman source)

- **`expose:` → `ExposeHostPort=` is correct, not a semantic inversion.** Per `podman-systemd.unit(5)` (*"Equivalent to the Podman `--expose` option"*) and `podman-run(1)` (*"`--expose`… has no effect on the actual networking rules unless `-P/--publish-all` is used"*). Quadlet `ExposeHostPort=` is intra-container reachability. Podup's existing mapping at `internal/quadlet/unit/container.rs:302-303` is correct. **No fix needed here.**
- **CDI device names do not silently drop.** `ExtractCDIDevices` in `pkg/specgen/generate/container_create.go:281` uses `parser.IsQualifiedName` (CDI) to split CDI from non-CDI entries. Non-CDI entries stay in `s.Devices` and are passed to the runtime. `registry.InjectDevices` fails with a clear error if the CDI name is not in the registry. **No silent drop.**
- **`container_name` is already validated by regex.** `WithName` in `libpod/options.go:329` uses `types.NameRegex = "^[a-zA-Z0-9][a-zA-Z0-9_.-]*$"`. There is no reserved-name list against system services. The conflict real concern is with pods/containers existing on the host, which podup already disambiguates via the `podup.project` label.

## Findings still in scope (the actual work)

- **`label_file:` content unsanitised** — `internal/engine/container/fields.rs:163-195`. Per-line keys/values are split on `=` and inserted into `SpecGenerator.labels` with no control-character check. Podman JSON-encodes the value, so wire injection is safe; downstream consumers that re-parse the label value are the concern. **The fix is to reject ASCII control characters** (NUL, newlines) and cap label count and value length.
- **Per-line size cap on `label_file`** — same file, no cap on the resulting HashMap size, on key length, or on value length. A 16 MiB file with one-byte keys and many entries would produce millions of labels in the spec.
[...]

PR #1382, opened 2026-08-08
fix: sanitize label_file content and URL-encode the podup.config-files label (#1361)

Jaro-c:
The label_file reader split each line on '=' and inserted the (key, value) pair into SpecGenerator.labels with no control-character or size check. Podman JSON-encodes the value on the wire, so wire injection was never the concern — the concern is downstream consumers that re-parse the label value (logs, ls/inspect UIs, label-based filters), where a control character in either side lets one entry break out of its context. The 16 MiB read cap on the file also left the resulting HashMap size, key length, and value length unconstrained.

This change adds `sanitize_kv_pair`, which validates a single key/value pair:

- key: non-empty, no ASCII control characters, length ≤ 253 bytes (Podman's per-label cap);
- value: no ASCII control characters, length ≤ 4 KiB;
- map: capped at 64 distinct entries, with overwrites of an existing key still allowed at the cap (the cap bounds distinct keys, not insertions).

`build_label_file_labels` now returns `Result<HashMap<String, String>>` so a rejection surfaces as an `Unsupported` error naming the file and line, rather than a silently-truncated map.

The `podup.config-files` label joins compose-file paths with `,` and so could visually merge with the next entry when a path itself contained a `,`. Each path is now passed through `encode_path_for_label` which replaces `,` with `%2C`, making the round-trip through any downstream `,`-split unambiguous.

## Test plan

- Unit tests: `sanitize_kv_pair` — bad input (control chars, oversize key/value, too many entries) → reject, good input → pass.
- Integration: a 1000-line label file → capped at 64 entries; a label file with control characters → reject at parse with a clear error; a path containing `,` in `compose_files` → URL-encoded in the `podup.config-files` label.
- Live-Podman: the lane (Podman 5 + 6).


Jaro-c:
Labels: `type:security`, `prio:P3`, `effort:M`, `area:engine`, `status:ready`

Jaro-c:
_(removed: that comment was just labelling the PR; the labels are now set on the PR itself)_

Issue #1365, opened 2026-08-07
fix(libpod): bound stream allocations and classify parse errors before reporting


Jaro-c:
## What's wrong

The libpod stream handling has four concrete correctness gaps. None of them is a panic; they are DoS surfaces and error-classification holes. Validation against the libpod server source (`containers/podman/pkg/api/handlers/compat/containers_logs.go:121-126`, `containers_stats.go:67-89`, `events.go:55-89`) **corrected the wire format**: the frame header is **8 bytes** (not 4), and **stats and events are NDJSON, not frame-stream** — they need a different parser from logs.

## What the libpod server actually sends

- **Logs** (`/containers/{name}/logs?follow=true`): 8-byte frame header — 4 bytes stream type (1=stdout, 2=stderr) + 4 bytes size in big-endian uint32 — then the frame body. Format matches Docker: <https://docs.docker.com/reference/api/engine/version/v1.46/#tag/Container/operation/ContainerLogs>. Wire-format ceiling: 4 GiB per frame (`uint32` max).
- **Stats** (`/containers/{name}/stats?stream=true`): NDJSON. Each sample is one JSON object encoded with `coder.Encode(jsonOut)` + flush. No frame header. Client splits on newline.
- **Events** (`/events?stream=true`): NDJSON. Same pattern as stats.
- **Server cap**: none. The server writes what the runtime produces; no maximum frame size on logs, no maximum sample size on stats/events. The cap is the client's responsibility.

## Gaps in scope (corrected from the original)

- **Logs: cap before the allocation.** The current code (`internal/libpod/types/stream.rs:52-66`) reads the 4-byte size from `header[4:8]`, allocates `8 + size` in one shot, then checks `buf.len() > MAX_STREAM_BUF`. A hostile or buggy server sending `size = 0xFFFFFFFF` triggers a 4 GiB allocation, then the cap fires. **Cap must be enforced BEFORE the allocation.**
- **Logs: reasonable cap value.** 1 MiB matches moby (`MaxFrameSize` in `libcontainerd`). The current `MAX_STREAM_BUF` is undocumented; pick a concrete value.
- **Stats and events: cumulative cap.** The streaming cap checks `buf.len() > MAX_STREAM_BUF` after `buf.extend_from_slice(&data)` (parsers at `stream.rs:108-110, 169-172`). A server that emits many small frames whose `buf.len()` always dips below the cap between reads can still pin the process. **Add a `u64 total_received` counter** alongside the per-buffer cap.
[...]

Issue #1362, opened 2026-08-07
perf(client): introduce HTTP connection reuse in the libpod client


Jaro-c:
## What's wrong

`Client::connect` (`internal/libpod/client/mod.rs:124-132`) opens a brand-new hyper HTTP/1.1 connection over the Unix socket for *every* request. The doc comment justifies this as "correct for a CLI tool where API calls are sequential and infrequent", but `up` is fan-out by construction: a 100-service compose file issues ~5 libpod calls per service (prefetch presence probe, container-list, create, start, healthcheck) plus a per-level container-list bulk — easily **600+ connections per `up`**.

Validation against the libpod server source (`containers/podman/pkg/api/server/server.go`) confirmed the diagnosis but corrected the rationale: **the server DOES allow connection reuse**. Go's `net/http` over a Unix socket is standard keep-alive; the server's `IdleTimeout: opts.Timeout * 2` applies equally to reused and fresh connections, and there is no `MaxConcurrentStreams` cap. The bug is the **client side** opening a fresh connection per request, not the server refusing reuse.

## Scope

- A connection pool keyed by socket path, holding N hyper HTTP/1.1 connections. Acquire on request, release on completion. Per-socket health check (drop on `ECONNRESET`, `Broken pipe`).
- The streaming endpoints (`get_stream`, `post_json_stream`) take a connection for the lifetime of the stream and release on stream end. They do not share with the buffered pool.
- The 64 MiB body cap (`MAX_RESPONSE_BYTES`) and the per-frame stream cap (`MAX_STREAM_BUF`) stay where they are; the pool is orthogonal.
- A `--connection-pool-size` (or env `PODUP_LIBCOD_POOL`) knob for users who want to tune; default 8 connections.
- Tests: sequential calls reuse one socket (asserted via a counter on `tokio::net::UnixStream::connect`); parallel calls spread across N sockets; a dropped connection is replaced transparently; a closed pool on process exit doesn't leak file descriptors.

## Test plan

- **Unit test** with a fake-libpod server that counts incoming connections. Sequential 100 requests → 1 connection. Concurrent 16 requests → ≤16 connections. A forcibly-closed connection is replaced on next request.
- **Benchmark** (separate): a 50-service `up` with and without the pool, on the same VM, three runs each. Median + p95. The bench harness is the right place for this; not a CI check.
- **Lane**: real-Podman 5 and 6, no regression.

## Static-review limit

[...]

Issue #1322, opened 2026-08-03
The integration suite is load-flaky on both Podman majors, and not only under nested virt


Jaro-c:
Measured on 2026-08-03 while verifying #1270, on hardware the product context does not currently account for.

## What the context says today

> A subset of archive/streaming tests still flakes with `IncompleteMessage` (connection drops) **under nested-virt**.

The nested-virt attribution is what the lane sees, and it is not the whole story.

## What I measured

**Podman 6.0.1, a plain KVM guest — not nested virt.** Full `engine_integration` suite, same VM, same podman build, minutes apart:

| tree | result |
|---|---|
| `develop` + the #1270 fix | 170 passed, **8 failed** |
| `develop` unchanged | 169 passed, **9 failed** |

The two failing sets overlap heavily but are not identical, which is the signature of load rather than of a defect in either tree. Every failure is an `IncompleteMessage` or `unexpected EOF during chunk size line`, and they land on `engine.up()` — not on the archive path #1270 was about.

**Podman 5.7.0, the owner host.** One full-suite run failed `watch_sync_and_restart_does_both` with `sync+restart did not copy the file`. That same test then passed **6 of 6 in isolation, on both trees**. Earlier the same day another full-suite run reported 2 failures whose identities I did not capture, against ~121 s for a clean run and 230 s for that one.

## Why this matters more than the count

The testing standard is unambiguous: *a flaky test is a P1 bug — fix it or delete it, never retry-loop around it.* Right now the suite is retry-looped around by hand, by whoever re-runs it and sees green.

It also costs real diagnostic time. Verifying #1270 needed a control run on the unmodified tree specifically because a red suite could not be read as evidence on its own, and one failure had to be chased with six isolated re-runs before it could be called noise.

And it interacts with the lane gates: Podman 6 has **no** known-failures list and is held only by a pass-count floor of 115 (`.github/podman-baseline-tests`), which is exactly the shape #1039 chose because the failing set carries a variable tail. That variable tail is this.

## What would move it

Not a retry, and not raising the floor.

- **Bound the concurrency.** `cargo test` runs the integration tests in parallel against one podman socket. Whether the drops track the number of in-flight requests is a one-flag experiment (`--test-threads`), and it is the cheapest thing that could turn "flaky" into "a documented limit".
- **Separate transport failure from test failure.** #1104 is the open question underneath — podup cannot tell a finished libpod stream from a broken one — and every failure here is that same ambiguity arriving through a different door.
[...]

PR #1385, opened 2026-08-08
fix: surface libpod field names in 4xx/5xx errors (#1357)

### LINKED ISSUE #1357: fix(security): validate compose fields forwarded verbatim to libpod
## What's wrong

podup forwards several compose fields to libpod with no shape check. The proposed fix was a per-field allow-list. After validating against the podman source, that framing is wrong: libpod **already** validates where it matters (namespaces, device_cgroup_rule), and the rest are **intentionally open** by design — `cap_add`, `extra_hosts`, `runtime`, `devices`, `cgroup_parent`, `ipam.options` are user-supplied strings whose failure surfaces from the runtime or the cgroup manager, not from libpod's specgen. A podup-side allow-list duplicates libpod's parser in another language and gets out of date when libpod's parser changes.

The right fix is **better surfacing of libpod's error responses**: when libpod returns a 4xx/5xx with a Podman-shaped error, podup should surface it as a compose-shaped error with the offending field named. Today, the 4xx/5xx path falls into a `PodmanError::Api` and the user sees the raw Podman message.

## What libpod actually validates (from `containers/podman` source)

- `pid` / `ipc` / `uts` / `cgroup` → `ParseNamespace` in `pkg/specgen/namespaces.go` accepts only `host | private | pod | ns:<path> | container:<id>` (plus `shareable`/`none` for ipc). **Already validated.**
- `device_cgroup_rules` → `parseLinuxResourcesDeviceAccess` in `pkg/specgenutil/specgen.go:514` validates `type ∈ {a,b,c}`, `major`/`minor` numeric or `*`, `mode ∈ {r,w,m}`. **Already validated.**
- `cap_add` / `cap_drop` → `pkg/specgenutil/specgen.go:436-441` copies the slice verbatim. Failure is at the OCI runtime, not in libpod. `ALL` is a real special value (man `--cap-drop.md`).
- `extra_hosts` → `--add-host.md` documents `hostname[;hostname[;...]]:ip` with `;` as the multi-hostname separator and `host-gateway` as a keyword. Parsing is `strings.Cut` on `:`. A podup-side allow-list contradicts the format.
- `runtime` → `--runtime.md` documents *"The path to an alternate OCI-compatible runtime"*. The value is a path; libpod does not pre-verify the path exists.
- `devices` → `ExtractCDIDevices` in `pkg/specgen/generate/container_create.go:281` uses `parser.IsQualifiedName` (CDI) to split CDI from non-CDI entries. Non-CDI entries stay in `spec.Devices` and are passed to the runtime.
- `cgroup_parent` → `WithCgroupParent` only rejects empty string. Path normalization is the cgroup manager's job.
- `ipam.options` → passed to the IPAM driver as `map[string]string`. No set of "known" keys.

## What podup's "validator" should actually do

Surface libpod's error better. When libpod returns 400/500, the current `PodmanError::Api` propagates the message but loses the field name. Two specific improvements:

1. When the request body is a `SpecGenerator` (`internal/libpod/types/container/spec/`), capture the field name from the failing path (e.g. `pid: "host"` is fine; `pid: "evil"` should surface as `service.web: pid namespace "evil" not recognised`).
2. For build queries (build args, labels, annotations, target platforms) — the same pattern: a 4xx with an opaque message becomes a clear "field X has value Y, which the runtime rejected for reason Z" error.

## Test plan

- **Integration test**: a compose with `cap_add: [CAP_TOTALLY_MADE_UP]`, `extra_hosts: ["db:127.0.0.1,github.com:127.0.0.1"]`, `runtime: "/nonexistent"`, `devices: ["/dev/sda"]`, `pid: "host"` (legit), `pid: "evil"` (invalid). For each, assert that podup produces a clear, field-named error, not a raw libpod HTTP body.
- **Live-Podman**: the lane (Podman 5 + 6) on this PR.
- **No new CI** beyond the existing lane; this is a UX change, not a security boundary.

## Out of scope

- Host-binding/privilege warnings (separate issue, #1358).
- Path/value sanitization (separate issue, #1361).
- The `podup` field-name parsing improvements above are independent of this issue; they are podup-internal error-rendering work, not podman validation.

Jaro-c:
## What this fixes

When libpod returns a 4xx/5xx, the operator used to see the raw libpod message with no compose-side context. This change pre-validates the fields libpod validates on its own and surfaces a compose-shaped error: `service.<name>: <field>: ... (value: <offending value>)`. For the cases where the engine knows the field, a new `post_json_with_field` / `check_status_with_field` path promotes a generic libpod 4xx to the same field-shaped form.

Fixes #1357.

## What changed

- **`internal/libpod/error.rs`** — new `PodmanError::Field` variant carrying `{service, field, value, message}`. `Display` leads with the compose-side service and field, keeps the libpod message in `message` so the cause is not lost, and truncates/escapes the value so a huge or binary value cannot flood the rendered error.
- **`internal/libpod/validate.rs`** (new) — pure pre-validators that mirror libpod's allow-lists (no podup-side duplication of OCI runtime checks, netns modes, etc.):
  - namespace modes (`pid` / `ipc` / `uts` / `cgroup` / `userns_mode`) against libpod's `ParseNamespace` allow-list;
  - `device_cgroup_rule` access strings against `parseLinuxResourcesDeviceAccess` (`r`, `w`, `m` only);
  - `build.args` / `build.labels` keys against the buildkit parser's `[A-Za-z0-9_.-]+` charset.
- **`internal/libpod/client/mod.rs`** — extracted `parse_error_message` from `check_status` so the existing helper is shared; added `check_status_with_field` and the public `post_json_with_field` / `post_json_ok_with_field` methods, which promote a 4xx/5xx to a `Field` error when the caller knows the field. Without a context the existing `Api` shape is preserved — the new methods are purely additive.
- **`internal/engine/container/mod.rs`** — `pre_validate_spec` runs after `device_cgroup_rule` is built but before the `SpecGenerator` is assembled, so a rejected value fails the call before any POST to the daemon.
- **`internal/engine/build/mod.rs`** — `pre_validate_build` runs after `build_args` and labels are resolved but before the build URL is assembled, so a bad key fails the call before any POST.
- **`tests/engine_integration/error_surfacing.rs`** (new) — integration tests that run against real Podman (skip when unreachable) and assert the field-named error shape for `pid: evil`, `ipc: bogus`, malformed `build.args` keys, `cap_add: [CAP_TOTALLY_MADE_UP]`, `runtime: /nonexistent/runtime`, and `extra_hosts: [bad]`. The `pid: host` happy path is also pinned so a future regression that over-validates is caught.

## Out of scope (intentional)

The fields libpod forwards verbatim — `cap_add`, `runtime`, `devices`, `extra_hosts` — deliberately do **not** get a podup-side allow-list. Their failure surfaces from the OCI runtime or cgroup manager, not libpod's specgen, and a podup-side allow-list would duplicate the OCI runtime's parser in another language and drift out of date when libpod or runc changes. The integration tests assert the libpod message is preserved and the surfaced error is operator-readable (no raw libpod JSON body) for those cases.

## Validation

- `cargo build` (lib + bin + tests with `--features test-helpers`): clean.
- `cargo test --lib`: 1591 passed, 0 failed.
- `cargo test --test engine_integration --features test-helpers error_surfacing`: 8 passed, 0 failed (real Podman 5.x on the runner).
- `cargo fmt --check`: clean.
- `cargo clippy --all-targets --features test-helpers -- -D warnings`: clean.
- Lane (Podman 5 + 6) will exercise the integration suite on this PR.

## Issue status

This PR does **not** close #1357 — the issue stays open until the lane goes green and the fix ships on a release.


Jaro-c:
Closing as deferred. The line-cap fail on mod.rs (627 code lines after #1357) cannot be cleanly resolved by splitting the file — the impl Client block is shared with prior work (#1362 connection pool, #1365 stream bounds) and the call-graph between send/read_body/the public API methods does not split across files without breaking visibility. The work from #1357 itself is correct; the line-cap is an orthogonal refactor tracked as a separate issue. Re-done as a clean follow-up PR if needed.

Jaro-c:
Reopening. I closed this as deferred because `internal/libpod/client/mod.rs` came out at 627 code lines and the `impl Client` block could not be split cleanly, since it was shared with the connection pool (#1362) and the stream bounds (#1365).

That blocker is gone. #1390 split the client by HTTP verb, so `get.rs`, `post.rs`, `put.rs` and `delete.rs` are separate files now and there is somewhere for this work to land. I merged current `develop` into the branch and moved `post_json_with_field` into `post.rs` where the rest of the POST path lives. `check_status_with_field` stays in `mod.rs` because it is not verb-specific and `post.rs` calls it from there.

Measured on the merged branch:

| Check | Result |
|---|---|
| `internal/libpod/client/mod.rs` | 273 code lines (hard limit 500, was 627 when I closed this) |
| `cargo check --all-features` | passes |
| `cargo test --lib` | 1638 passed, 0 failed |
| `cargo test --bins` | 87 passed, 0 failed |
| `cargo test --test engine_integration -- --test-threads=2` | 187 passed, 0 failed, against Podman 5.7.0 |
| `tests/engine_integration/error_surfacing.rs` | 8 passed, 0 failed |
| `cargo fmt --check` / `cargo clippy --all-features` | clean |

The integration run is capped at two threads on purpose. The suite drops connections above three (#1322), so an uncapped local red says nothing.

I also checked the new code is actually reachable rather than merged-but-orphaned: `pre_validate_spec` is wired in `internal/engine/container/mod.rs`, `pre_validate_build` in `internal/engine/build/mod.rs`, and `spec_field_error` in `internal/libpod/client/mod.rs`.

PR #1431, opened 2026-08-17
ci(release): verify every file the installers fetch is staged

Jaro-c:
Consumer side of Glyndor/.github#133. **Draft on purpose — see Blocked below.**

## Why

The asset-contract gate compares `install.sh` against `release.yml` and answers *"do the two agree on the asset names"*. It cannot answer *"does the release produce them"* — a workflow file is a statement of intent, not proof that a step ran.

`Glyndor/.github` carried a check that tried to answer it anyway, by looking for `SHA256SUMS` among the parsed `asset:` keys. Those keys are the build matrix, and the manifest is computed *from* the matrix outputs in a later step, so it can never match. **That is the false positive on #1425** — the gate fails this repo for doing it correctly.

## What this does

Adds one step to the `release` job, between "Verify every signature against the keys consumers embed" and "Create GitHub Release". At that point everything is built, signed and checksummed, and the release is not immutable yet — so existence is a stat call rather than an inference.

What it requires is derived from `install.sh`'s own `${BASE_URL}` fetches, not hardcoded. That matters: **the old check never looked at `SHA256SUMS.sig`**, which `install.sh` downloads and without which it aborts. A release missing only the signature would have passed a gate whose stated purpose was making downloads verifiable.

## Verified, and not

Run locally against this repo's real `install.sh`, `install.ps1` and `release.yml`:

| case | result |
|---|---|
| complete staging dir | `OK: all 8 files the install scripts fetch are staged` |
| `SHA256SUMS.sig` removed | fails — the case the old check could not see |
| `SHA256SUMS` removed | fails, naming both missing files |
| the amended gate, against this repo | passes (it did not before) |

**Not verified:** the workflow wiring itself. `release.yml` only runs on a tag push, and dispatching it would cut a real release, so the step order and the `upstream/` checkout path are unexercised until the next release actually runs. I am not claiming otherwise.

## Blocked

The `ref:` currently pins the branch SHA of Glyndor/.github#133. That SHA becomes unreachable once #133 is squash-merged, so **this must not merge as it stands**. Order:

1. Glyndor/.github#133 merges.
2. `.github` cuts a tag.
3. #1425 bumps the `installer-contract` pin to that tag — that run is what proves the gate fix green.
4. This PR's `ref:` moves to the tag, and comes out of draft.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -453,7 +453,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: Glyndor/.github
-          ref: 118244b8d758bdec6d2c5115d815dbc2e5c4f4e9 # main (pre-tag)
+          ref: 118244b8d758bdec6d2c5115d815dbc2e5c4f4e9 # v1.16.0
           persist-credentials: false
           path: upstream
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -453,7 +453,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: Glyndor/.github
-          ref: e429f62795b09d17452c6d45db8018390d606483 # ci/installer-contract-verify-substance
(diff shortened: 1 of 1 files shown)


Jaro-c:
Glyndor/.github#133 is merged, so the branch SHA this pinned is gone. Repointed at `118244b` on `main`.

Still draft: the `ref:` should be a tag, not a branch commit, before this merges. Waiting on the `.github` tag.

Issue #1500, opened 2026-08-21
ComposeError::Include is a dead variant: nothing constructs it


Jaro-c:
`ComposeError::Include(String)` is public API that no code path produces. Found while writing coverage for the error variants nothing was exercising (#1499).

## Measured

Every occurrence in the tree:

| where | what |
|---|---|
| `internal/error.rs:51` | the definition |
| `internal/error.rs:227` | its `Display` arm |
| `internal/error.rs:392` | a `Display` test that builds it by hand |

That is all of them. No production path constructs it, so **an `include:` that fails does not surface as `Include`** — it comes out as `Parse`, or `FileNotFound`, or nothing at all depending on where it enters.

## Why it matters beyond tidiness

The variant is public and the enum is `#[non_exhaustive]`, so a consumer can `match` on `Include` today and write a handler that will never run. Worse, the handler *looks* correct in review: the variant exists, the arm compiles, and nothing indicates it is unreachable.

## Options

1. **Wire it up** — have whatever resolves `include:` emit `Include(...)` on failure. This is the option that makes the type honest, but it changes what users see for a failing `include:`, so it wants a look at what each entry point currently returns.
2. **Remove it** — breaking for anyone matching on it, which is what a major is for.
3. **Leave it** — with a doc comment saying it is reserved, so the next reader does not assume it is reachable.

I would take (1): the variant names a real failure mode and the alternative today is an error that does not say `include:` was involved. But that is a call about behaviour, not a cleanup, which is why it is filed rather than folded into the coverage PR.

Jaro-c:
Fixed on `develop` by #1511.

Closing by hand rather than waiting: `Closes #N` only fires on a merge into the default branch, and in this repository the fix lands on `develop` first, so nothing was going to close this on its own.

PR #1521, opened 2026-08-21
docs: document the 35 undocumented public items and deny the 36th

### LINKED ISSUE #1520: 35 public items have no doc comment, and nothing stops the 36th
The style standard says every public item carries a doc comment addressed to a third-party
reader. Nothing enforces it, and `rustdoc` is the surface helmly-agent and epistle read.

Measured by turning the lint on and counting: **35 items**, no more.

```
28 missing documentation for a struct field
 7 missing documentation for a variant
```

All in six files:

| file | items |
|---|---|
| `internal/error.rs` | 16 — the fields and variants of `ComposeError` |
| `internal/ui/progress/board.rs` | 8 — `Kind`'s five variants and `Row`'s three fields |
| `internal/libpod/error.rs` | 4 — `Api { status, message }`, `IncompatibleApiVersion` |
| `internal/libpod/types/stream.rs` | 2 — `StdOut`/`StdErr` payloads |
| `internal/ui/progress/live.rs` | 2 — `Stdout`/`Stderr` |
| `internal/compose/types/develop.rs` | 2 — `watch`, `unknown` |

Sixteen of the thirty-five are `ComposeError`'s own fields. That is the type a library consumer
matches on, and #1478 and #1500 were both about making that surface usable — a consumer who
can now name `ComposeError::Include` still cannot read what is inside it without opening the
source.

## The ask

Document the 35, then turn on `#![deny(missing_docs)]` in `internal/lib.rs` beside the
existing `#![deny(unsafe_code)]`, so the next undocumented public item fails the build instead
of being noticed a year later.

Doc comments state the contract — what the field means, its units, its invariants — not how
it is produced, and never a restatement of the name. `path: String` documented as "the path"
is worse than nothing, because it satisfies the lint and tells the reader less than the field
name already did.

## Why it is worth doing now rather than later

The count only goes up, and `deny` is cheap to add while the list is 35 and expensive once it
is 300. It is also the natural companion to the API work already merged this cycle: the
predicates are public, the error variants are reachable, and the last missing piece is that
`cargo doc` says what any of it means.

Jaro-c:
The style standard says every public item carries a doc comment addressed to a third-party
reader. Nothing enforced it, and `rustdoc` is the surface helmly-agent and epistle read.

Turning the lint on counted the gap exactly: **35 items**, 28 struct fields and 7 variants,
across six files.

| file | items |
|---|---|
| `internal/error.rs` | 16 — `ComposeError`'s fields |
| `internal/ui/progress/board.rs` | 8 — `Kind`'s variants, `Row`'s fields |
| `internal/libpod/error.rs` | 4 |
| `internal/libpod/types/stream.rs` | 2 |
| `internal/ui/progress/live.rs` | 2 |
| `internal/compose/types/develop.rs` | 2 |

Sixteen of the thirty-five are `ComposeError`'s own fields — the type a library consumer
matches on, and the one #1478 and #1500 spent this cycle making reachable. A consumer who can
now name `ComposeError::Include` still could not read what is inside it without opening the
source.

## The comments state a contract, not the field name again

`path: String` documented as "the path" satisfies the lint and tells the reader less than the
field name already did. So each one says something the signature does not:

- which identifier a field carries — `WaitServiceExited::container` is the Podman container
  name with its replica suffix, `ScalePortConflict::service` is the compose key;
- what a value excludes — `ScalePortConflict::ports` lists only fixed host ports, because a
  range or an unspecified port cannot collide that way;
- what a reader must not assume — a `LogOutput` frame boundary is **not** a line boundary, so
  one frame can split a line;
- where the wording comes from — `Api::message` is libpod's own text, preserved verbatim, so
  a caller matching on it is matching on libpod's wording rather than ours;
- that `develop.watch` rules are order-sensitive, first match wins.

## The gate

`#![deny(missing_docs)]` goes in beside the existing `#![deny(unsafe_code)]`, with a comment
saying why now: cheap to adopt at 35, expensive at three hundred, and the count only rises.

## Test plan

- `cargo check --lib`, `cargo check --lib --all-features`, `cargo check --bins`: all clean.
- `cargo test --lib error::`: 32 passed.
- **Verified the gate fires rather than assuming it.** A lint nobody has watched fail is
  decoration, so I deleted one of the new comments and re-ran:
  ```
  error: missing documentation for a struct field
  error: could not compile `podup` (lib) due to 1 previous error
  ```
  Restored: clean.

Closes #1520

### DIFF
--- a/internal/compose/types/develop.rs
+++ b/internal/compose/types/develop.rs
@@ -10,6 +10,8 @@ use serde::{Deserialize, Serialize};
 /// `develop:` service key — holds file-watch rules for the `watch` command.
 #[derive(Debug, Clone, Deserialize, Serialize, Default)]
 pub struct DevelopConfig {
+	/// The rules in declaration order. `watch` evaluates them in order and the
+	/// first matching path wins, so order is meaningful rather than cosmetic.
 	#[serde(default, skip_serializing_if = "Vec::is_empty")]
 	pub watch: Vec<WatchRule>,
 }
@@ -36,6 +38,9 @@ pub struct WatchRule {
 	/// Command run inside the container for the `sync+exec` action.
 	#[serde(skip_serializing_if = "Option::is_none")]
 	pub exec: Option<WatchExec>,
+	/// Keys under this rule that podup does not recognise, kept so `config`
+	/// round-trips a compose file without silently dropping them. Insertion
+	/// order is preserved.
 	#[serde(flatten, default, skip_serializing_if = "indexmap::IndexMap::is_empty")]
 	pub unknown: indexmap::IndexMap<String, serde_yaml::Value>,
 }
--- a/internal/error.rs
+++ b/internal/error.rs
@@ -27,7 +27,14 @@ pub enum ComposeError {
 	/// A service has neither an `image:` nor a `build:` section.
 	NoImageOrBuild(String),
 	/// A `${VAR}` with the `?err` modifier was required but unset.
-	RequiredVarNotSet { var: String, msg: String },
+	RequiredVarNotSet {
+		/// The variable name as it appeared in the compose file, without the
+		/// `${}` or the `:?` suffix.
+		var: String,
+		/// The author's own explanation, taken verbatim from the text after
+		/// `:?` in the interpolation. Empty when the form was bare `${VAR?}`.
+		msg: String,
+	},
 	/// A `${…}` interpolation reference is malformed (e.g. an invalid character
 	/// in the variable name, as in `${FOO BAR}` or `${FOO.BAR}`).
 	InvalidSubstitution(String),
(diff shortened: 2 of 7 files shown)


Jaro-c:
Rebased onto `develop` at `2830d0c`, which now carries #1511, #1514 and #1516.

Worth saying why rather than just doing it: `deny(missing_docs)` is a crate-wide lint, so it is one of the few changes where every pull request can be green individually and the merged result still fails — any public item added by another branch in the same window lands undocumented into a tree that now refuses it. Per-PR checks cannot see that.

Checked it directly rather than assuming: `cargo check --lib --all-features` and `cargo check --bins --tests` are both clean against the rebased tree. None of the three merges added a public item.

The rebase also clears the advisory `dependabot-freshness` red this branch was carrying from having been cut before #1505.

Jaro-c:
Checked the remaining ordering risk too, since `deny(missing_docs)` is the one change here that can be green per pull request and still break on merge.

Merged this branch and #1522 together onto `develop` locally and built the result:

- `cargo check --lib --all-features`: clean
- `cargo check --tests --bins`: clean
- `cargo test --lib`: 1710 passed, 0 failed

So neither merge order is a problem. #1522's new items in `internal/engine/copy/stream.rs` are `pub(super)`, which the lint does not reach, and `extract_tar_guarded` keeps its doc comment through the signature change — it is public under `--all-features` via `fuzz_api`, so that one did need checking rather than assuming.

PR #1525, opened 2026-08-21
fix(debian): build the package against musl so it installs on every supported release

### LINKED ISSUE #1487: The .deb is built by a compiler CI never runs, in an unpinned sid
I hit this while packaging epistle and went looking at how podup does it, so this is a report from the outside rather than a change — the call is yours.

Two things about `build-deb` in `release.yml`, both measured today.

**The .deb is built by a compiler CI never exercises.** `ci.yml` calls `rust-ci.yml` without overriding `toolchain`, so it validates on the reusable's pinned default of **1.97**. `build-deb` runs in `container: debian:sid` and builds with Debian's packaged `rustc`, which is **1.95** right now. So the binary users install through apt comes from a compiler nothing in CI has ever run the suite against. It is a small gap while the two versions are close, and it is not bounded by anything — sid moves on its own schedule.

**`debian:sid` is an unpinned moving tag.** Two releases of byte-identical source can be built in different environments with nothing in the history saying so. The CI standard is explicit about the shape of this ("Pin the toolchain, don't track `latest`/`stable`", with the Rust 1.97 incident behind it), and while that clause is written about reusables installing `stable`, a `container:` that resolves to whatever Debian unstable holds that morning is the same failure mode with a different spelling.

Related, and the reason I looked at all: **`debian/control` declares `rustc (>= 1.85)`, and that floor is derived from `rust-version` rather than measured.** It happens to be true for podup today. It was false for epistle — `edition = "2024"` implies 1.85 and trixie ships exactly 1.85, but `rcgen`, `time` and `time-macros` each require 1.88 and the build died on a resolution nothing in the manifest mentions. Worth re-measuring podup's floor by actually building in trixie rather than trusting the number, because the day a dependency raises it, the failure lands at tag time on an immutable release.

For what it is worth, what epistle ended up with after the owner picked between the options:

- `release.yml` builds in the official `rust:1.97.1-slim-trixie` image, digest-pinned, so the shipped artifact comes from the compiler that validated the code.
- A separate `debian.yml` on every PR builds through the shared `rust-debian.yml` reusable with Debian's own rustc on a digest-pinned `debian:sid`, with `dpkg-checkbuilddeps` running — which is what proves `Build-Depends` is satisfiable rather than aspirational.

Neither build subsumes the other, and splitting them is what made both questions answerable. Whether that shape suits podup is a different question — podup ships five targets and a much wider matrix, so the cost is not the same.

Refs Glyndor/epistle#705.

### LINKED ISSUE #1512: apt install fails on Debian 12 and Ubuntu 22.04: the .deb needs glibc 2.39
`apt install podup` cannot work on Debian 12 or Ubuntu 22.04. Read straight out of the live
archive index:

```
$ curl -sS https://apt.glyndor.net/dists/stable/main/binary-amd64/Packages
Package: podup
Version: 3.8.0
Depends: libc6 (>= 2.39), libgcc-s1 (>= 4.2)

Package: epistle
Version: 0.3.5
Depends: libc6 (>= 2.39), libgcc-s1 (>= 4.2)
```

Identical on `binary-arm64`. Both products, both architectures.

What each distribution ships, read from the distributors' own package indexes rather than
from memory:

| distribution | `libc6` | `>= 2.39` |
|---|---|---|
| Debian 12 bookworm | 2.36 | **no** |
| Ubuntu 22.04 jammy | 2.35 | **no** |
| Ubuntu 24.04 noble | 2.39 | yes, exactly at the floor |
| Debian 13 trixie | 2.41 | yes |

Debian 12 is oldstable with security support into 2028 and Ubuntu 22.04 is an LTS supported
into 2027. On either one, adding the repository succeeds, `apt update` succeeds, and
`apt install podup` fails on an unmet dependency. Nothing upstream of that point warns.

## It also falsifies a documented design decision

The archive is a single `Suites: stable` for every Debian and Ubuntu, and the private context
gives the reason in as many words: the `.deb`s are *"static, self-contained binaries (no
system-library deps …), so they install identically on Debian 12, Ubuntu 22.04/24.04, etc."*,
with "keeping binaries static is the requirement that keeps the single suite valid."

They are not static. They link `libc6` and `libgcc-s1` dynamically, with a floor. So the
premise the single-suite design rests on does not hold, and the two distributions named in
that sentence as working are the two that do not.

## Where the floor comes from

`release.yml`'s `build-deb` job runs in `container: debian:sid` and installs Debian's `cargo`
and `rustc`. The floor is not chosen anywhere — it is whatever glibc symbol versions the
binary happens to reference, resolved against whatever glibc `sid` carried on the day of the
build. A moving tag decides which distributions can install the product.

That is the same line of YAML as #1487, which covers the *compiler* half (the `.deb` is built
by a rustc the suite never runs against). This is the *libc* half, and it is the one users
feel. Filed separately because the fixes differ and this one is user-facing.

## Options, none of them free

1. **Build in the oldest distribution to support** (`debian:bookworm`) instead of `sid`. The
   floor drops to 2.36 and both LTS releases come back. Cost: bookworm's rustc is far below
   the crate's floor, so the toolchain would have to come from rustup inside the container,
   which is the decision #1487 is about.
2. **Build against musl** and make the claim true. Genuinely static, no libc floor at all.
   Cost: a different target, and musl's allocator is measurably slower unless swapped.
3. **Keep the floor and say so.** Declare the supported set in the README and the archive
   landing page, and stop claiming one suite serves every Debian and Ubuntu. Cheapest, and
   honest, but it drops two supported LTS releases from the product.

I lean to (1) — it restores the documented behaviour rather than redefining it — but it is
coupled to #1487 and should be decided with it.

## Not a defect, checked while measuring this

The archive serves `podup 3.8.0` while `v4.0.0` is the latest release. That is the documented
≤24h pickup: `InRelease` is dated 20 Aug 18:25 UTC, `v4.0.0` published at 19:57 UTC, and the
publish cron runs at 06:00. The next run carries it. Recording it so nobody else re-measures
it as a symptom.

Jaro-c:
Every published `.deb` declared `Depends: libc6 (>= 2.39)`. Debian 12 ships 2.36 and Ubuntu
22.04 ships 2.35, so `apt install` failed on an oldstable supported into 2028 and an LTS into
2027 — after adding the repository and `apt update` had both succeeded, with nothing upstream
warning.

Nobody chose that floor. The Linux **release binaries have always been static musl**; the
`.deb` was the one Linux artifact that was not, because `dpkg-buildpackage` rebuilt from source
with Debian's `rustc` against whatever glibc the build container carried.

## Changes

- **`debian/rules`** builds for a musl target derived from `DEB_HOST_ARCH`, and **errors**
  rather than falling back to the host. A package that quietly returned to glibc would
  reintroduce the floor with nothing to notice it, on a release that is immutable.
- **`release.yml`** installs rustup pinned to `1.97` — the same version the `build` job and
  `rust-ci.yml` use — plus `musl-tools`, and drops Debian's `cargo`/`rustc`. Debian's
  toolchain ships no musl standard library, so `--target` would not resolve. That closes the
  other half of the same line of YAML (**#1487**): the packaged binary was compiled by a
  compiler the suite had never run against.
- **`debian/control`** drops `cargo, rustc` from `Build-Depends` rather than declaring a
  compiler that no longer does the work — `dpkg-checkbuilddeps` would otherwise pass on the
  wrong thing, which *is* the defect. `Depends` needs no change at all: it is generated from
  `${shlibs:Depends}`.

## Measured, not assumed

**The combination nobody had built** — musl *with* the package's trimmed feature set — on the
pinned toolchain:

```
$ cargo +1.97 build --release --locked --target x86_64-unknown-linux-musl \
      --no-default-features --features watch,completions
    Finished `release` profile [optimized] target(s) in 1m 26s

$ file target/x86_64-unknown-linux-musl/release/podup
ELF 64-bit LSB pie executable, x86-64, static-pie linked, stripped
$ ldd  target/.../podup      →  statically linked
$ .../podup --version        →  podup version v4.0.0
$ .../podup completions bash →  emits the script (this is what dh_auto_install runs)
```

**That `${shlibs:Depends}` really does empty out**, rather than reasoning it:

```
$ dpkg-shlibdeps -O .scratch/podup
dpkg-shlibdeps: warning: cannot parse dynamic symbol definition: no symbols
(exit 0, and no dependency line emitted at all)
```

**Both architectures and the failure path**, via `make -n`:

```
DEB_HOST_ARCH=amd64   → cargo build … --target x86_64-unknown-linux-musl …
DEB_HOST_ARCH=arm64   → cargo build … --target aarch64-unknown-linux-musl …
DEB_HOST_ARCH=riscv64 → debian/rules:27: *** no musl Rust target is mapped for
                        DEB_HOST_ARCH=riscv64; add one above.  Stop.
```

`install` and the three `completions` invocations all resolve to
`target/<triple>/release/podup`.

Also checked, because it is not obvious: **`dpkg` accepts `#` comment lines in `debian/control`** —
`dpkg-checkbuilddeps` parses the file and reports only the genuinely unmet
`debhelper-compat`, not a syntax complaint. And `musl-tools` turns out not to be needed for the
compile itself (the Rust musl target links self-contained here), but it stays for the same
reason the `build` job installs it.

## What is unverified

**I did not run a full `dpkg-buildpackage`.** Everything above is the pieces; the assertion
that matters is the `Depends:` line on a real package:

```
dpkg-deb -f podup_*.deb Depends     # must not name libc6
```

That is what `debian / dpkg-buildpackage` and `debian / build (vendored / offline)` run on this
pull request, and they are the gate. If either comes back with `libc6` still there, the
`dpkg-shlibdeps` result above did not carry into the real build and this is not done.

## Scope against #1487

Closes the toolchain half — the `.deb` is now built by the pinned compiler CI validates with.
**It does not close #1487**, which also covers `container: debian:sid` being a moving tag. I
left that alone: Dependabot's `github-actions` ecosystem does not update a workflow's
`container:` image, so a digest pin has no bump path and a frozen `sid` eventually stops
resolving `apt-get install` — the answer there is probably a stable release rather than a
digest, and that is a separate decision.

Closes #1512


Jaro-c:
Re-pointed the test at the new subject.

Replaces `tests/debian_rustc_floor.rs` with `tests/workflow_toolchain_pin.rs` (8ba212c). The `Build-Depends: rustc (>= X)` line that file asserted against is gone with this branch — the build is now driven by the toolchain the workflows pin. The invariant moves with it.

The new file holds three tests:

- `the_two_workflow_pins_declare_the_same_toolchain` reads `release.yml`'s `build-deb` job `RUST_TOOLCHAIN` env and `debian-build.yml`'s `rust-toolchain` input, and asserts they are equal.
- `the_pinned_toolchain_satisfies_the_crate_floor` reads the same two plus `Cargo.toml`'s `rust-version`, asserts the pins still agree, and that the agreed value satisfies the crate floor — compared as version numbers (`"1.9"` must sort below `"1.85"`, not by string).
- `the_parsers_read_the_values_rather_than_guessing_them` pins the parsers on inputs that differ from today's files, including a fixture where `build`/`build-deb`/`release` all carry different toolchains, a comment that mentions `RUST_TOOLCHAIN:` with a fake value, and `cmp_version` cases covering `"1.9"` vs `"1.85"`, `1.85.0` vs `1.85`, and cross-major.

Mutation checks (both restored):

- build-deb set to `1.98` with `debian-build.yml` still at `1.97`: `the_two_workflow_pins_declare_the_same_toolchain` and `the_pinned_toolchain_satisfies_the_crate_floor` both fail at the agreement `assert_eq!`, message points at #1487.
- both pins set to `1.80` (still agreeing, below `rust-version = "1.85"`): the agreement test stays green, `the_pinned_toolchain_satisfies_the_crate_floor` fails at the `assert_ne!` against `Ordering::Less`, with the "builds with a compiler that does not satisfy the crate" message.

Nothing else on the branch touched. `debian/rules`, `debian/control` and the workflow edits are unchanged from the merge base.

Issue #1517, opened 2026-08-21
The identity-colour registry is process-wide, so a second project evicts the first


Jaro-c:
The identity-colour registry is a process-wide static:

```rust
// internal/ui/mod.rs:112
static SERVICES: std::sync::RwLock<Option<HashMap<String, usize>>> =
    std::sync::RwLock::new(None);

pub fn set_services(names: &[String]) {
    if let Ok(mut slot) = SERVICES.write() {
        *slot = Some(palette::assign(names));   // replaces, never merges
    }
}
```

`set_services` **replaces** the map. Its own doc comment says *"Call once per invocation"*,
which is true for the CLI and false for the thing this crate also is: a library.

So with two `Engine` values alive in one process — which is exactly helmly-agent's shape,
managing a fleet of projects — the second `set_services` evicts the first project's names.
Every service of the first project then falls through to the hash fallback in `colour_for`,
and its colours change underneath a run that is still going. Nothing errors; the log just
starts colouring the same service differently.

`service_slot` reads the same static, so there is no per-`Engine` path at all.

## Why the fallback does not save it

The hash fallback is good design for its stated purpose — an orphan container, a project from
`ls`, anything podup renders but did not resolve from a compose file. A stable wrong colour
beats no colour. What it cannot do is keep the *registered* answer stable, and
`colour_for_prefers_the_registered_slot_over_the_hash` already pins that the two disagree.
That test is what proves the eviction is observable rather than cosmetic.

## Scope

This is the state I could find. `set_project` next to it has the same shape. I have not
audited the rest of `internal/ui/` for other process-wide statics, and the fix should start
by doing that rather than by patching this one.

## Options

1. **Per-`Engine` registry.** The honest fix. `Engine` carries the map and the renderers take
   it, so two engines cannot see each other's. Touches the signature of everything that
   colours a label.
2. **Merge instead of replace.** One line, keeps the static, and two projects with a
   colliding service name still fight over the slot — it narrows the window without closing
   it.
3. **Document the constraint and enforce it.** Say plainly that the crate supports one
   project per process, and make a second `set_services` with different names loud rather
   than silent.

(1) is the one that matches what the library claims to be. It is a real piece of work and
wants its own design pass, which is why this is filed rather than folded into #1502.

#1502 covers the missing test only; this covers the defect that test exposes.

Jaro-c:
Did the audit this issue said it needed, and it narrows the scope rather than widening it.

[...]

Issue #1367, opened 2026-08-07
fix: TOCTOU and resource leaks in cp, run_attached, install, and secrets paths


Jaro-c:
## What's wrong

Four concrete TOCTOU / resource-leak issues in the engine, the install path, and the secret creation path. The project lock closes the cross-invocation case for some; static review cannot tell whether a within-invocation race is reachable.

## Findings in scope

- **`cp_to_container` does two HEAD requests with a PUT between them** — `internal/engine/copy.rs:161-220`. A concurrent container mutation between the first and second HEAD can land a successful-but-wrong-state PUT. The per-project lock closes the cross-invocation case; the within-invocation case is closed by libpod's own validation. Re-check the contract; document the assumption.
- **`run_attached` opens the hijack before start, drops on start failure** — `internal/engine/lifecycle/run_attached.rs:28-49`. If `post_empty_ok(start_path)` returns `Err`, the socket's kernel buffer and the bytes the container may have already written are dropped without drain. The success path relies on the `pump_terminal` drain; the failure path does not. Drain on both paths.
- **`Engine::new` silently falls back to empty `PathBuf` when `current_dir()` fails** — `internal/engine/mod.rs:151`. A user whose working directory was deleted between `Engine::new` and any path-resolution call gets an empty `base_dir` and a confusing "compose file not found" later. Document the contract, or use `Engine::with_base_dir` (the safe constructor) from the CLI.
- **`create_secret` inspect-then-delete-then-create** — `internal/engine/secrets/create.rs:103-171`. The per-project lock closes the cross-invocation case. The within-invocation case is closed by the unique project-scoped naming. Re-document the assumption; the lock and the naming are the only two things keeping this safe.
- **`install_binary` reads the current binary with `.ok()`** — `internal/update/install.rs:73-90`. If the read fails (permission denied, target deleted), the in-memory backup is `None`, the new bytes overwrite the target, the self-test fails, and the operator gets "no backup was available to roll back" — but the new bytes are on disk and the old ones are gone. The Windows path's `target.old` survives a kill; the Unix path does not.
- **`install_at` Windows path uses `std::fs::File::create` without `O_NOFOLLOW`** — `internal/update/install.rs:271-274`. The Unix path uses `O_NOFOLLOW | O_EXCL`; the Windows path follows symlinks. On a shared install directory an attacker can plant a symlink at `tmp` and redirect the install.

## Scope

- `cp_to_container`: document the lock + libpod-validation contract in a comment at the top of the function.
[...]

PR #1531, opened 2026-08-22
fix(ui): key the identity-colour registry by project (#1517)

Jaro-c:
Two `Engine` values alive in one process (helmly-agent's fleet shape)
saw the second `set_services` evict the first project's colours. The
engine itself writes the registry from `lifecycle/down_label.rs:60`, so
an embedder could not avoid it by discipline.

The registry is now keyed by project name. `set_services` reads the
project set by `set_project` and inserts under that key; `identity_slot`
and `service_slot` iterate every registered project before falling back
to the per-label hash. The engine call site in `down_by_label.rs` now
calls `set_project` before its `set_services`, so the label-only
teardown keys under its own project. `PROJECT_PREFIX` stays as a
fallback strip so `every_spelling_of_one_container_gets_one_colour`
still pins the no-registration collapse to one hash slot.

The seven other process-wide statics this audit looked at are correct
and say so in their own doc comments — the progress trio because
`begin` is inert without the CLI, the signal-handler latch because it
already names the embedding-crate case, `quadlet`'s `NO_WARN` because it
is a `thread_local!`, and `palette`'s `AVAILABLE` because it is a
property of the process's stdout.

## Design

The honest fix would be an `Engine`-owned registry threaded through
every renderer, but `service_slot` is read from deep inside — the log
prefixer, `progress_line`, the table renderers, `stats`, `events`,
`paint_status_cell` — so threading would cross several layer boundaries.
I rejected that option because the public API of `ui` cannot widen
(#1478, #1481 made that surface deliberate).

The third option, merging registrations rather than replacing, narrows
the window without closing it: two projects sharing a service name
(`web`, `db`, `api`) still collide. I rejected that because helmly-agent's
fleet would hit it on day one.

Keying the existing `SERVICES` map by project — option 2 — is the same
answer for the registry shape (one process, N slot tables) without
threading anything new. The only signature change is adding
`set_project` before `set_services` in the engine path; the public API
stays exactly the same.

The bare-key fallback (`identity_slot` and `service_slot`) iterates
every project's services in HashMap order, first match wins. In a
single-project process this is identical to the legacy behaviour. In
helmly-agent's fleet it is the linear-in-projects lookup the design
admits; the alternative would be a thread-local active engine, which
does not survive across the async boundaries the engine's own call
sites cross.

## Tests

Inverts `a_second_project_evicts_the_first_projects_colours` (#1518)
to assert `after == registered` (the first project's slot survives a
second registration), keeping the same fixture names and the
discriminating `assert_ne!(registered, unregistered)` guard. Adds a
`set_project` to `set_services_makes_registered_names_distinct` for
the same reason — `set_services` is now project-keyed, so a registration
without a project name is a no-op.

Mutation check: I temporarily changed `set_services` to clear the
project map on every call (the original eviction behaviour) and the
inverted test went red on the surviving-slot assertion. Restored;
all 1710 lib tests pass with `--all-features`.

## Breaking?

No. `set_project`, `set_services`, `identity_style`, and `service_style`
keep their signatures. The behaviour change is observable only across
the multi-engine boundary the issue pins. `set_services` is now a no-op
without a prior `set_project` for the same project; every existing
caller (CLI commands and the engine path, post-fix) already calls
`set_project` first.

### DIFF
--- a/internal/engine/lifecycle/down_label.rs
+++ b/internal/engine/lifecycle/down_label.rs
@@ -54,9 +54,16 @@ impl Engine {
 		// Without this every container here fell back to the per-label hash
 		// instead of the sequential identity colour `ps` and the compose-file
 		// `down` path use — the same class of defect fixed for `logs` (#1082).
+		//
+		// `set_services` registers under the project's name in the process-wide
+		// identity-colour registry. Two `Engine` values alive in one process
+		// (helmly-agent's fleet) key off their own `self.project`, so neither
+		// registration evicts the other (#1517). `set_project` must come
+		// first; without it the registration is a no-op.
 		let by_service = self.live_project_replicas().await?;
 		let mut service_names: Vec<String> = by_service.keys().cloned().collect();
 		service_names.sort();
+		crate::ui::set_project(&self.project);
 		crate::ui::set_services(&service_names);
 
 		let containers: Vec<String> = by_service.into_values().flatten().collect();
--- a/internal/ui/mod.rs
+++ b/internal/ui/mod.rs
@@ -85,11 +85,18 @@ static PROJECT: std::sync::RwLock<String> = std::sync::RwLock::new(String::new()
 
 /// The pre-formatted `"{project}-"` prefix, cached so [`identity_slot`] does
 /// not reallocate it on every call. Set in tandem with [`PROJECT`] by
-/// [`set_project`]; read by [`identity_slot`] on every progress event, table
-/// row, and log-prefix row (~300 of those per 100-service `up`, #1364).
+/// [`set_project`]; read by [`identity_slot`]'s no-registration fallback path
+/// so `proj-web-1` and `web-1` still collapse to one hash slot when no
+/// project has called `set_services` yet. The main registered lookup iterates
+/// the per-project prefix entries in [`SERVICES`] directly, so this cache
+/// only matters when nothing has been registered.
 static PROJECT_PREFIX: std::sync::RwLock<String> = std::sync::RwLock::new(String::new());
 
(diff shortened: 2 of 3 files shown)


Jaro-c:
Reviewed and pushed two changes on top. The design choice is right and the reasoning for rejecting the other two holds — option 1 really does cross the log prefixer, `progress_line`, the table renderers, `stats` and `events`, and option 3 leaves two projects sharing a service name (`web`, `db`, `api`) still colliding, which is helmly-agent's fleet on day one.

## The silent return

`set_services` returned quietly when no project was set. That is the failure shape this repository hunts: **colours still appear** — every label drops to the hash fallback — so nothing looks broken and nothing says why. The CLI always calls `set_project` first, so it never surfaces there; a library consumer has no such habit, and that is exactly who the change is for.

Now it emits a `tracing::warn!` naming the missing call. Same treatment `Engine::new` got in #1367 for its silent empty-`PathBuf` fallback.

## The test name

It kept `a_second_project_evicts_the_first_projects_colours` while its assertion was inverted to check the opposite. That is not cosmetic here: this repository has already paid for it. The three build tests using `$VAR` inside `dockerfile_inline` were *not* weak tests — they verified a different mechanism than the one they named, and could not notice. A name that contradicts the assertion is the same defect with the two halves swapped.

Renamed to `a_second_project_leaves_the_first_projects_colours_alone`.

## Independently verified

I re-ran the mutation rather than taking the report: restoring the eviction (`map.clear()` before `insert`) turns the test red; restoring gives **75 passed** in `ui::`, and `cargo check --tests --all-features` is clean.

Agreed on **not breaking**: the public signatures are unchanged and the behaviour difference is only observable across the multi-engine boundary the issue pins.

PR #1535, opened 2026-08-22
ci(debian): prove the new Depends gate fires (do not merge)

Jaro-c:
Temporary proof branch for the gate added in Glyndor/.github#141. Not for merging — I will close it once both directions are on record.

The reusable now fails a musl build whose package declares a shared-library dependency. That predicate was checked in a shell against six inputs, but the integration was not: whether `RUST_TARGET` reaches the confirm step, whether `dpkg-deb -f` returns what the pattern expects, and whether the `case` matches at all.

So this branch pins `debian-build.yml` to the pull request's branch and changes one line of `debian/rules` to build `x86_64-unknown-linux-gnu` while the workflow input still says `x86_64-unknown-linux-musl`. That is the drift the gate exists to catch: the declared target and the built binary disagree, every other check stays green, and today nothing notices.

Expected: `debian / dpkg-buildpackage` fails, naming the `libc6` dependency. I will then revert only the `debian/rules` line and confirm the same pin goes green.

A gate nobody has watched fail is decoration.

### DIFF
--- a/.github/workflows/debian-build.yml
+++ b/.github/workflows/debian-build.yml
@@ -59,7 +59,7 @@ jobs:
     # proven on a real consumer before its tag is cut - .github's own CI never
     # exercises a caller. This repository's dpkg-buildpackage job was that
     # proof, going from red to green in 6m36s on #1525.
-    uses: Glyndor/.github/.github/workflows/rust-debian.yml@e70ff47fd8aefa0f054846815f19b98768d61122 # v1.17.0
+    uses: Glyndor/.github/.github/workflows/rust-debian.yml@4b8df6cc119877ba55452b6fd7d9b5ef8eab09ef # TEMP: branch of Glyndor/.github#141
     with:
       package-name: podup
       check-vendored: true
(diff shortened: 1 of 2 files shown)


Jaro-c:
Both directions are on record now, with the same reusable pin and only `debian/rules` differing between the two runs.

Fails, run 32555320707:

    ##[error]built for x86_64-unknown-linux-musl but the package declares a
    shared-library dependency: libc6 (>= 2.39), libgcc-s1 (>= 4.2)

That is Glyndor/podup#1512's exact floor, caught at build time instead of by a user whose apt refuses to install.

Passes, run 32555607152: green, with only the `debian/rules` line reverted.

One correction worth recording. My first attempt at the mutation also failed, and it failed for the wrong reason — I put the comment on the assignment line, and make keeps the space before a trailing `#` inside the value, so `RUST_TARGET` came out as `"x86_64-unknown-linux-gnu "` and `install` died before any `.deb` existed. The gate never ran. Had I stopped at "it failed, so it works", I would have signed off a check on no evidence at all. A fixture has to be valid in every respect except the one under test.

Closing unmerged — this branch only ever existed to produce those two runs.

Issue #1526, opened 2026-08-21
Eleven pinned tool versions, nothing watches any of them


Jaro-c:
The decision was to stop using whatever Rust the distribution shipped and pin the latest
ourselves. The producing half of that — pin, do not track `stable` — is done and correct; an
unpinned toolchain upgraded itself in July and turned every open pull request here red on a
clippy lint nobody introduced.

**The consuming half does not exist.** The CI standard says to bump it "through a new
`.github` release, so each repo adopts a new toolchain deliberately", and nothing proposes
that bump. Eleven versions are pinned by hand across the reusables and **nothing watches any
of them**.

## Measured on the packaging path

| piece | pinned | latest | watched by |
|---|---|---|---|
| Rust toolchain | `1.97` | **1.98.0** (2026-08-18) | nothing |
| `debhelper-compat` | `13` | 14, sid only | nothing |
| `Standards-Version` | `4.7.0` | **4.7.4** | nothing |
| release container | `debian:sid`, unpinned | — | nothing |
| CI container | `debian:trixie@sha256:…` | — | pinned by digest |
| Cargo dependencies | — | 12 updates available | **Dependabot, daily** |

Dependencies are covered. What is not covered is every version a human typed into a workflow.

Only Rust was actually measured against upstream. The other ten are unmeasured, which is the
finding rather than an omission in it.

## Three constraints, so nobody bumps these in one sweep

1. **The `.deb`'s toolchain must equal `rust-ci`'s.** Raising only the package's reintroduces
   #1487 — a binary compiled by a compiler the suite never ran against. `rust-ci` moves first,
   org-wide, and the package follows through the `rust-toolchain` input that
   `Glyndor/.github#138` adds.
2. **`debhelper-compat 14` exists only in sid.** CI builds in `trixie`, still on 13. Raising it
   today breaks the CI build.
3. **The 12 Cargo updates include `rustls` and `rustls-webpki`, and none of them reach the
   `.deb`.** It builds with `--no-default-features --features watch,completions`, which drops
   the TLS stack. They affect the release binaries and `install.sh`, not the package.

## What this issue covers

`Standards-Version` to 4.7.4 (cheap, no dependencies), and `debhelper-compat` to 14 **once
trixie carries it** — or once someone decides CI should build in sid, which is a different
question and belongs with #1487's container discussion.

The Rust bump and the watcher live in `Glyndor/.github`, which has issues disabled, so their
problem statement travels in their pull request bodies per the workflow standard's exception.

[...]

PR #1558, opened 2026-08-25
ci: watch the pinned tool versions nothing else watches

Jaro-c:
Closes the watcher half of #1526.

## Why now

The issue asked for this — "the watcher, because otherwise this recurs" — and one pin has already recurred. The `debian:trixie` digest was pinned in #1530 and has drifted since, exactly as `release.yml`'s own comment predicts: Dependabot's `github-actions` ecosystem does not update a workflow's `container:` image, so a digest pin has no bump path.

## What it covers

The four pins this repository owns that nothing else watches. Rust is covered by `rust-toolchain-watch` in Glyndor/.github and Podman by `podman-watch`, so neither is repeated.

Weekly at 05:00 Monday — off `audit.yml`'s 06:00 and `fuzz.yml`'s 07:00, since stacking crons on one minute only invites scheduling delays. One issue listing every drift, not one per pin, deduplicated on a title that never changes. Registered with `schedule-freshness` at 15 days so a cron that stops firing is itself reported.

## Verified by running it, not by reading it

The step's script was extracted from the workflow and executed against the real repository files with a stubbed `gh`:

```
| Pin                            | Pinned          | Upstream        |
| debian:trixie container digest | sha256:fac46bff… | sha256:34cd9e9f… |
| python-version                 | 3.13            | 3.14            |
```

Both match what I measured by hand beforehand. Just as important is what it did **not** report: `Standards-Version` (4.7.4.1, current) and `debhelper-compat` (13, which is trixie's major) stayed silent. The debhelper check reading sid instead of trixie was the trap most likely to be built in, and it would have filed "bump to 14" every Monday — a demand that breaks the build if followed.

The deduplication path was exercised too: with the issue already open, the run reports `Drift found, but the issue is already open — nothing to do.` and creates nothing.

`actionlint` is clean **with shellcheck enabled**. That matters here: an earlier draft passed `actionlint -shellcheck=` while containing a heredoc whose terminator was indented, so it never closed. Running the script is what found it, along with two more:

- a `python3 -c` block written across several indented lines. It works on this machine (Python 3.14) and fails on the runner (3.12, `IndentationError`) — confirmed against a real 3.12 container, not assumed from the changelog.
- the cron I first specified collided with `fuzz.yml`'s.

## Not fixing the digest here

I measured that it is stale but am not bumping it blind. `release.yml` only runs on a release, so no pull-request CI would exercise the new image. The watcher will file it, and it should be raised where a build can prove it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,14 +106,13 @@ jobs:
       workflow: podman-watch.yml
       max-age-days: 3
 
-  freshness-pin-watch:
-    permissions:
-      contents: read
-      actions: read
-    uses: Glyndor/.github/.github/workflows/schedule-freshness.yml@76f8bc6dd79baa995d5bfa347c4b985433cb0ce1 # v1.20.0
-    with:
-      workflow: pin-watch.yml
-      max-age-days: 15
+  # A newly added scheduled workflow cannot be registered here in the same pull
+  # request that adds it. The guard passes only on a run with `event=schedule`
+  # and `status=success`; GitHub fires crons solely from the default branch, and
+  # `workflow_dispatch` does not count. So a fresh entry fails with "No
+  # successful scheduled run on record" until the workflow has reached `main`
+  # and its cron has fired once. `pin-watch.yml` (#1526) is waiting for exactly
+  # that and gets its entry afterwards.
 
   # Monthly (`0 4 1 * *`), so seventeen days without a run is normal and a
   # weekly threshold here would fail on ordinary work for most of the month.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,15 @@ jobs:
       workflow: podman-watch.yml
       max-age-days: 3
 
+  freshness-pin-watch:
+    permissions:
+      contents: read
+      actions: read
+    uses: Glyndor/.github/.github/workflows/schedule-freshness.yml@76f8bc6dd79baa995d5bfa347c4b985433cb0ce1 # v1.20.0
+    with:
+      workflow: pin-watch.yml
+      max-age-days: 15
+
   # Monthly (`0 4 1 * *`), so seventeen days without a run is normal and a
   # weekly threshold here would fail on ordinary work for most of the month.
   freshness-benchmark:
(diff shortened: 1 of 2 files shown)


Jaro-c:
The `freshness-pin-watch` check failed, and it was right to.

`schedule-freshness` passes only on a run with `event=schedule&status=success`. GitHub fires crons solely from the default branch, and `workflow_dispatch` does not count. A workflow added in the same pull request that registers it therefore has no such run, and the guard reports "No successful scheduled run on record" — the guard working as designed, not a bug in it.

The four entries that pass all have their workflow on `main` already; `pin-watch.yml` does not yet. Its registration comes after it reaches `main` and its cron fires once, which for a weekly schedule is up to seven days after the next release.

Removed the entry from this pull request and left the reason as a comment beside the others, where whoever adds the next scheduled workflow will hit the same wall.

Worth noting the ordering, since it is not obvious: a new scheduled workflow needs two merges — one to land it, one to register it — and they cannot be the same one.

Jaro-c:
Correcting a claim in this pull request's description. One of the three defects I reported was not a defect.

**What I claimed:** that the first draft contained a heredoc whose terminator was indented and therefore never closed, and that running the script is what found it.

**What is actually true:** it closed fine. Inside a YAML block scalar (`run: |`), YAML strips the block's indentation before bash ever sees the script, so a heredoc whose opener and terminator sit at the *same* indentation both arrive at column 0. Reconstructed the original and ran it through `yaml.safe_load` the way the runner does:

```
PY_OUT=$(python3 - <<'PY'
...
PY
)
→ resultado: 1
  3.13
```

Correct output.

**Why I got it wrong:** I tested the heredoc by pasting the indented text into `bash -c`, which skips YAML's dedent entirely. So I measured a script the runner would never see, and reported the result as a confirmed defect. Same failure shape as reading a truncating regex's output as the value — the instrument was wrong, and the answer it gave was confident.

**What stands:** the other two findings were real.

- The indented `python3 -c` source genuinely fails on the runner. Automatic dedent of `-c` arrived in CPython 3.13; `ubuntu-24.04` ships 3.12. Verified in a real `python:3.12-slim` container, which raises `IndentationError: unexpected indent`. That one is not affected by YAML dedent, because the source is inside a shell double-quoted string rather than a heredoc body.
- The cron collided with `fuzz.yml`'s `0 7 * * 1`.

**Does the merged code need changing?** No. What landed avoids multi-line Python entirely — plain `grep`/`sort -u` for the python-version check and one-line `python3 -c` elsewhere — and it was verified end to end against the real repository files, finding the two real drifts and correctly staying silent on the two that are current. It is fine, and arguably simpler. But it was not fixing what I said it was fixing.

PR #1576, opened 2026-08-28
chore(release): 5.2.0

Jaro-c:
Two behavioural fixes reach users with this one.

push --ignore-push-failures returned 0 when every push failed, so a gate written
as `podup push --ignore-push-failures && deploy.sh` deployed against a registry
that had received nothing. Only every-attempt-failed is an error now; one
survivor still succeeds, which is what the flag is for.

stats --format json swallowed serialisation failures, printing an empty line and
exiting 0 where the other five list commands propagate.

Minor rather than patch: both change observable exit codes, and a script that
depended on the old zero will notice. Not major -- neither is a deliberate API
change, they are corrections of behaviour that was wrong.

All three version files bumped together: Cargo.toml, Cargo.lock and
debian/changelog. The release workflow's verify job compares them before
building, which is the gate 1.9.1 added after 1.9.0 shipped .debs labelled
1.8.0.

### DIFF
--- a/.github/workflows/asset-contract.yml
+++ b/.github/workflows/asset-contract.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v6.0.0
         with:
-          python-version: "3.13"
+          python-version: "3.14"
       - name: Signing requirements must parse as separate pinned entries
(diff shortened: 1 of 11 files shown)


Jaro-c:
Wrong source branch: main-guard requires pull requests into `main` to come from `develop`, and this came from `release/5.2.0`.

Reopening as develop -> main once the bump lands on develop, which is the flow the standard describes.

Issue #1583, opened 2026-08-28
The debian-build lane and the release build the .deb in different trixie snapshots


Jaro-c:
Two workflows build the `.deb` and they build it in different images.

| where | `debian:trixie` digest | is it the current trixie tag? |
| --- | --- | --- |
| `release.yml:239` (`container:`) | `sha256:f324c7ff54321e8d9c588493a20244965938ce0aa50bbd1022d38010e9ffc4b1` | yes |
| `reusable-rust-debian.yml:24` (`debian-image` default) | `sha256:fac46bff2e02f51425b6e33b0e1169f55dfb053d83511ca28aa50c09fd5ed7a4` | no, it is behind |

`debian-build.yml` is the lane that runs on every pull request, and it takes the reusable's default. So the package the suite exercises comes out of a snapshot that is no longer what the release is built in, which is the opposite of what `release.yml`'s own comment claimed: "the release that ships to users was produced by the same image the suite exercises".

They were also already different before #1568 replaced the remote reference. `Glyndor/.github`'s `rust-debian.yml` carries the same `fac46b` default, so the divergence is not something the copy introduced; the copy only removed the last place anyone might have noticed.

Two things follow and they are separate calls:

- **The lane digest should move to the current trixie.** That is the one-line change, and it means the pull-request lane starts building in the image the release uses.
- **Whether the two should be one value at all** is the design question. A `container:` needs a literal, so the release workflow cannot read the reusable's default, and any answer here leaves two copies of one datum. Naming the second copy in the first one's comment is what this repository does elsewhere; a check is better, and there is one now.

I found this by running the new fifth pin in `pin-watch.yml` while working on #1582, which is also why the report arrived with the watcher's own title. #1582 extends pin 1 to read both sites and report the disagreement, so this stays visible every Monday until it is closed. It deliberately does not choose which digest wins, because that changes what gets built.

Jaro-c:
Landed on `develop` as `7022135` (#1587), squashed. Closing by hand, since a `Closes` trailer does not fire on a merge into `develop`.

The lane moves to `f324c7`, which is what the registry serves for `debian:trixie` and what `v5.2.1` was built in. The proof that it works is not a green tick on the pull request but the lane itself: `debian / dpkg-buildpackage` ran and its log shows `debian:trixie@sha256:f324c7ff5432`, so the pull-request lane has now built the package in the image the release is built in, which is the thing that had never happened.

[...]

Issue #1582, opened 2026-08-28
Finish the .github separation: the Rust pin is unwatched, five reusables sit outside their caller's path filter, and six comments went false


Jaro-c:
#1568 stopped this repository calling `Glyndor/.github` and I checked what that left behind. There is no `uses: Glyndor/` line on `develop` or on `main`, so the workflow half landed. What did not land is everything the copy could not carry, and one of the three is a real loss of coverage rather than a stale sentence.

## 1. The Rust toolchain pin is watched by nobody

`pin-watch.yml:9` says:

```
# Rust is watched by Glyndor/.github's rust-toolchain-watch.yml and Podman by
# this repository's podman-watch.yml, so neither is repeated here.
```

That sentence is the reason `pin-watch.yml` owns four pins instead of five. It is no longer true. `rust-toolchain-watch.yml` is `on: workflow_call` only, so it does nothing until a consumer calls it from its own scheduled workflow, and its own header says so. Nothing here calls it, and it was not among the reusables copied across. I searched every repository in the org and the only reference outside `Glyndor/.github` itself is that comment.

So the pin sits at `1.98` and nothing tells me when upstream moves. That is the exact failure the watcher exists to prevent: in 2026-07 Rust 1.97 landed, widened a clippy lint, and turned every open pull request here red on a change nobody made.

There is a second thing worth catching here that the reusable could not, because it takes a single `toolchain` input. The toolchain is pinned in **two** places now:

| file | value |
| --- | --- |
| `.github/workflows/reusable-rust-ci.yml` (`toolchain` default) | `1.98` |
| `.github/workflows/release.yml` (`RUST_TOOLCHAIN`) | `1.98` |

They agree today. Nothing makes them agree tomorrow, and a release built on a different toolchain than CI tested is the kind of difference that only shows up once, at the tag, on an immutable release.

## 2. Five path-filtered callers do not list the reusable they call

While the reusables lived in `Glyndor/.github`, a change to one reached this repository as a Dependabot bump that edited the caller file, which is inside the caller's own `paths:` filter. Now the reusable is a local file, and the filter does not name it:

| caller | reusable it calls, missing from its `paths:` |
| --- | --- |
| `asset-contract.yml` | `reusable-installer-contract.yml` |
| `debian-build.yml` | `reusable-rust-debian.yml` |
| `lint-powershell.yml` | `reusable-powershell-ci.yml` |
| `lint-shell.yml` | `reusable-shell-ci.yml` |
[...]

PR #1584, opened 2026-08-28
ci: finish the .github separation, and gate the three things the copy could not carry

Jaro-c:
## Summary

- Finishes what #1566 through #1571 started. There is no `uses: Glyndor/` line left on either branch, but three things the copy could not carry were still outstanding, and one of them was a loss of coverage rather than a stale sentence.
- The Rust toolchain pin had been watched by nobody since #1568, and the comment explaining its absence is what made that look deliberate.
- Five path-filtered workflows did not name the reusable they call, so a pull request editing only that reusable merged without running it.

## Changes

**The Rust pin is watched here now, and gated at pull-request time.** `pin-watch.yml:9` said Rust was covered by `Glyndor/.github`'s `rust-toolchain-watch.yml`. That reusable is `on: workflow_call` only, nothing in the org calls it, and it was not among the fourteen copied across. Copying it unchanged would not have worked either: it takes one `toolchain` input, and there is no single value here to hand it. The toolchain is written in **fifteen** places across eight workflow files, of which nine pin stable, four are pass-throughs, one is the fuzz nightly and one is a second copy of the MSRV.

So `pin-watch.yml` gains a fifth pin that reads all of them and compares the agreed value against upstream, and `tests/workflow_toolchain_pin/` grows from covering two sites to covering every one, on every pull request. The overlap between the two is deliberate and both files say so: the test answers within minutes of the edit and cannot reach the network, the watcher is the only thing comparing against upstream and runs whether or not the Rust suite did.

**A path-filtered workflow now has to name the reusable it calls.** Fixed in the five that did not, then made a rule in `reusable-workflow-lint.yml`, which reads every workflow on each pull request and sits outside the jobs it asserts about. Advisory, not required.

**The two `debian:trixie` digests disagree, and now something says so.** `release.yml` pins `f324c7`, which is the current trixie; `reusable-rust-debian.yml` defaults to `fac46b`, which is behind. The `debian-build` lane takes the default, so every pull request builds the `.deb` in a snapshot the release is not built in, which is the opposite of what `release.yml`'s comment claimed. That is #1583, and choosing which digest wins changes what gets built, so it is not decided here.

**Sixteen comments that the copy made false.** Ten input descriptions still said their version is "bumped via .github releases"; `reusable-workflow-lint`'s header described callers "that live outside this repository" and contrasted itself with an `actionlint.yml` that does not exist here; `reusable-shell-ci` justified a default with "the SHA pin is the buffer" after the pin had gone; `release.yml` pointed at `Glyndor/.github`'s `rust-ci.yml` as the file to stay in step with.

**Two defects that came out of running things rather than reading them.** `asset-contract` pins `1.85` as a second copy of the MSRV that `ci.yml` declares, with nothing keeping them in step. And piping `curl` into an `awk` that `exit`s on the first match closes the pipe underneath it, so under `pipefail` a fetch that succeeded reports 23 or 141; the upstream reusable has the same shape, which is consistent with it never having run.

**Three things that were checked and deliberately left alone.** `CODE_OF_CONDUCT.md`, `SECURITY.md`, `SUPPORT.md` and `FUNDING.yml` still come from the org `.github` repository, which is where the workflow standard puts them. `docs/security-model.md` linking to the org `SECURITY.md` is correct. And `CONTRIBUTING.md` recording where the reusables came from is provenance, not a dependency; it now names the tag, v1.21.0, and says byte-identity has stopped being a property to rely on, because eight of the fourteen already differ.

## Test plan

- [x] `cargo fmt --all --check` is clean
- [x] `cargo clippy --locked --all-targets --all-features -- -D warnings` is clean
- [x] `cargo test --locked --all-features --workspace` passes locally
- [x] `./tests/shell/*.test.sh` pass locally (every tracked test, including `ci-runs-every-test.test.sh`)
- [x] shellcheck is clean on every tracked `*.sh` and every extensionless script with a shell shebang. **At 0.10.0** locally; CI pins 0.11.0, and the two disagree about real constructs, so this claim is only as good as its version.
- [x] Podman lane ran, which I did not expect. I wrote that it would not because this branch touches no `internal/` code, and `podman-lane.yml` triggers on more than that. Both legs are green: `podman-vm (5)` in 17m28s and `podman-vm (6)` in 31m38s.
- [ ] Asset-contract fixtures ran. `asset-contract.yml`'s own filter now also matches `reusable-installer-contract.yml`, so this pull request triggers it.

**The integration suite needs its thread cap to be readable.** At full parallelism it reported two or three `build_images` failures, and the same thirteen tests pass run on their own. The `--test-threads=2` run the context prescribes came back `188 passed; 0 failed` in 896s on the same commit, so those reds are #1322's dose-response curve and not this change.

- [x] New or changed checks were verified by deleting the control and watching them fail

Every sabotage below was diffed against a copy taken beforehand and reported inert if it had not altered the file, and each was restored and re-run green afterwards.

`reusable-workflow-lint`'s new rule: removing the reusable line from `lint-shell` reports both its `push` and `pull_request` triggers by name; rewriting a `paths` block as `paths-ignore` reports the shape it does not model rather than passing over it; `.github/workflows/**` in place of the literal stays green; `**.sh` alone goes red.

`pin-watch`'s fifth pin: the nine stable sites disagreeing, the two MSRV sites disagreeing, the nightly exception rewritten to a beta channel, the MSRV exception rewritten to a nightly, and an unclassified new pin. The upstream comparison was exercised by moving all nine to 1.97. An unreadable digest in either file reports that file by name, which needed a `|| true` on the read: `grep` exiting 1 inside the substitution propagates under `set -e` and killed the run before the diagnostic, so the log said nothing about which pin it could not read.

`tests/workflow_toolchain_pin/every_site.rs`: one stable pin moved, the declared MSRV moved away from its copy, the nightly exception rewritten, an unclassified pin added, and the scanner itself narrowed to one key, which two tests catch, one of them the per-file blindness check. That check replaced a hardcoded count, because a total is a copy of the tree and goes stale the day a site is added.

**One thing to know about the new `reusable-workflow-lint` rule.** It has never run in CI. Promoting it to required needs the emitted name read off a real run, so it stays advisory.

## Checklist

- [x] Targets `develop`
- [x] Commits are signed off (DCO, `git commit -s`)
- [x] Commits are signed
- [x] Labels applied
- [x] Every new file in `tests/shell/` is named in some workflow's `test-command`. No new shell test here, and `ci-runs-every-test` was re-run after `lint-shell`'s `test-command` changed to `bash ./…`, plus driven red by dropping one suite from it
- [ ] Docs updated if behaviour changed. `CONTRIBUTING.md` is updated; no user-facing behaviour changed.
- [x] `Cargo.toml` and `debian/changelog` are at the same version. Both untouched.
- [ ] `cargo build --release` binary budget. No production code changed.

## Related issues

Refs #1582. Opened #1583 for the trixie digest divergence, which this pull request reports but deliberately does not resolve.

One thing I got wrong twice while writing #1582, corrected here: I said the toolchain lived in two places and that one description named `.github` releases. It is nine places and ten descriptions. Both numbers came from reading rather than counting, which is the same mistake the pins themselves are.


Jaro-c:
The capped run finished clean: `188 passed; 0 failed` in 896s at `--test-threads=2`, against the same commit that reported two or three `build_images` failures at full parallelism. So the reds were #1322's dose-response curve and nothing here, which was already the expectation given this branch touches no `internal/` code, but it is now measured rather than inferred from the isolation run.

Jaro-c:
Two follow-ups on this branch.

**The capped integration run finished clean:** `188 passed; 0 failed` in 896s at `--test-threads=2`, on the same commit that reported two or three `build_images` failures at full parallelism. So those reds were #1322's dose-response curve, now measured rather than inferred from the isolation run.

**`workflow-lint` was red and it was right.** The pin scanner skips pass-through values by testing for the expression opener, and I wrote that opener out. actionlint lexes it as an expression wherever it appears inside a `run:` block, whatever language the block is in, and reports the quote after it as a syntax error. Fixed by building the token from two halves next to the other constants.

Worth saying plainly: my local verification was parsing the YAML and running the extracted shell, and neither of those can see what actionlint makes of a `run:` block. actionlint v1.7.12, the version the reusable pins, runs here now and is clean over all fifteen workflows this branch touches. Bare `actionlint` walks the whole tree including `target/`, which is why it appeared to hang; scoped to the files, it is instant.

Issue #1553, opened 2026-08-24
Ubuntu 24.04 LTS cannot run podup: it ships Podman 4.9.3, and the README implies otherwise


Jaro-c:
podup requires **Podman >= 5.0**. Measured on repology 2026-08-24, this is what the distributions actually package:

| distribution | Podman | podup runs? |
|---|---|---|
| Debian 12 bookworm | 4.3.1 | no |
| **Debian 13 trixie** | **5.4.2** | yes |
| Debian 14 / unstable | 5.8.x | yes |
| Ubuntu 22.04 LTS | 3.4.4 | no |
| **Ubuntu 24.04 LTS** | **4.9.3** | **no** |
| Ubuntu 25.04 | 5.4.1 | yes |
| Ubuntu 25.10 | 5.4.2 | yes |
| Ubuntu 26.04 | 5.7.0 | yes |

**Ubuntu 24.04 is the LTS in support until 2029 and it does not reach 5.0.** `apt install podup` succeeds — the binary is static and installs fine — and then podup cannot talk to the engine.

## What the README says

> Many distributions still ship 4.x, so check `podman --version` and upgrade if needed. Fedora, Debian trixie/sid and recent Ubuntu releases carry 5.x

Every clause is true and the conclusion a reader draws is wrong. **"Recent Ubuntu releases" are the non-LTS ones.** The current LTS is not among them, and LTS is what someone administering a server is running — which is the reader this sentence is for.

## Two things to decide, and they are separate

**1. Documentation.** Name the LTS explicitly rather than saying "recent". Something a reader can check against their own `lsb_release` without having to know which Ubuntu releases are recent.

**2. Product.** Is Ubuntu 24.04 LTS supported or not? Today the answer is "no, silently". The options:

- document it as unsupported, and have the bootstrap installer say so before it installs anything rather than leaving a broken install behind
- point the installer at a Podman 5 source for that case (Podman upstream publishes packages; the Kubic/OBS repositories carry newer builds for Ubuntu)
- support Podman 4.x, which the version-gate rationale in the README argues against

## Note this is independent of the glibc floor

The two limits are different and only one is fixable by static linking:

| system | glibc >= 2.39 | Podman >= 5.0 |
|---|---|---|
| Debian 12 | fails | fails |
| Debian 13 | ok | ok |
| Ubuntu 22.04 | fails | fails |
| **Ubuntu 24.04** | **ok** | **fails** |

[...]

PR #1588, opened 2026-08-29
fix(packaging): require podman rather than recommending it

Jaro-c:
## Summary

- `podman (>= 5.0)` moves from `Recommends` to `Depends`. podup translates a compose file into libpod calls and does nothing else, so an install without an engine cannot do the thing it is for.
- A test ties the three places the Podman floor is written, because as a `Depends` that number decides whether the package installs at all.

## Changes

**The relationship.** `Recommends: podman (>= 5.0)` becomes part of `Depends`. The line arrived in the packaging skeleton in #36, when the goal was still the official Debian archive, and it is the only decision in `debian/control` with no reasoning beside it; the `Build-Depends` one carries a paragraph. It has one now.

**What changes for a user, stated rather than left to be discovered:**

| system | before | after |
|---|---|---|
| Debian trixie (Podman 5.4.2) | podman installed, because apt installs Recommends by default | podman installed, now as a requirement |
| Anything with `APT::Install-Recommends "false"` (minimal images, containers) | **neither engine nor warning** | podman installed |
| Ubuntu 24.04 LTS (Podman 4.9.3) | podup installs and cannot reach an engine | `apt install podup` **fails** |

The third row is the visible one. It is also the row the README has documented as unsupported since #1555, so apt now refuses what the documentation already said would not work, at install time instead of at first use.

**The case apt cannot express** is the operator driving a remote socket or a `podman machine`: a package relationship only sees the local machine, and there is no local engine to satisfy this. That case belongs to the release binary and `install.sh`, which warns rather than refusing when no local podman is present and takes `--skip-podman-check` when the local one is not the engine podup will use. The README says so beside the table now.

**`tests/podman_floor.rs`.** The floor is written in three places and derived from none of them:

- `internal/libpod/client/mod.rs` — `MIN_LIBPOD_API_MAJOR`, the gate the running binary applies
- `install.sh` — `PODMAN_MIN_MAJOR`, the installer precheck
- `debian/control` — the relationship

As a `Recommends` a wrong copy cost a suggestion. As a `Depends` a copy that drifts low lets apt install podup onto an engine the binary refuses, and one that drifts high refuses an install that would have worked. Neither surfaces until a user hits it, on a machine nobody here owns. The test also asserts the relationship is a `Depends` and not a `Recommends`, so putting it back is a red rather than a quiet regression.

`install.ps1` carries no floor and needs none: there is no Windows package relationship, and the engine lives in a `podman machine` VM whose version the installer cannot read from the host. That is written in the test rather than left as an absence.

## Test plan

- [x] `cargo fmt --all --check` is clean
- [x] `cargo clippy --locked --all-targets --all-features -- -D warnings` is clean
- [x] `cargo test --locked --all-features --workspace` passes locally
- [x] `./tests/shell/*.test.sh` pass locally
- [x] shellcheck is clean over every tracked `*.sh`, at 0.10.0; CI pins 0.11.0
- [ ] Podman lane ran. No `internal/` change.
- [x] Asset-contract fixtures ran. `install.sh` is unchanged here, but `debian/**` puts this in the debian lane's filter, which is the one that builds the package this changes.
- [x] New or changed checks were verified by deleting the control and watching them fail

Three sabotages, each diffed against a copy taken beforehand, and the restoration confirmed by reading `Depends:` and `PODMAN_MIN_MAJOR` back rather than by the test going green:

- the `debian/control` floor moved to `6.0`: `every_podman_floor_agrees` red, naming both values and what each direction costs;
- `PODMAN_MIN_MAJOR` moved to `4`: same test red, naming the installer;
- the relationship put back to `Recommends`: all three tests red, including the parser test, which is the shape that says the file no longer looks like what the parsers were written for.

**`dpkg-parsechangelog` and a hand parse both read the new stanza**, and the binary stanza carries exactly one relationship line mentioning podman.

One thing found on the way and **not** fixed here, because it is a different defect: the pre-push shellcheck command in `CONTRIBUTING.md` uses `find` over the working tree, so on a tree where a local `.deb` build has run it walks `debian/cargo-home/`, a gitignored vendored registry, and reports findings in third-party crates that CI never sees. `git ls-files -z '*.sh' '*.bash' | xargs -0` is clean on the same tree. Worth its own change.

## Checklist

- [x] Targets `develop`
- [x] Commits are signed off (DCO, `git commit -s`)
- [x] Commits are signed
- [x] Labels applied
- [x] Every new file in `tests/shell/` is named in some workflow's `test-command`. The new test is Rust, which cargo discovers.
- [x] Docs updated if behaviour changed. `README.md` and `docs/debian-packaging.md`, both of which described it as a recommendation.
- [x] `Cargo.toml` and `debian/changelog` are at the same version. Both 5.2.1, untouched; this is a packaging-relationship change and the next release carries it.
- [ ] Binary budget. No production code changed.

## Related issues

Refs #1553 and #1555, which established that Ubuntu 24.04 LTS is unsupported and that the product should say so before it breaks anything. This makes apt say it too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,10 @@ installs podup with apt — so upgrades and signing-key renewals arrive through
 packages. It leaves nothing of its own behind: the download is removed, and so
 is anything it had to install just to check the key.
 
-podup needs **Podman ≥ 5.0** (rootless) with its API socket listening. Podman is
-daemonless, but podup speaks the libpod API:
+podup needs **Podman ≥ 5.0** (rootless). The package depends on it, so apt
+installs it alongside podup, and refuses the install on a distribution whose
+Podman is older than that. Podman is daemonless, but podup speaks the libpod
+API, so the socket still has to be listening:
 
 ```sh
 systemctl --user enable --now podman.socket
@@ -105,9 +107,16 @@ it for its whole supported life.
 | Fedora 42 and newer           | 5.x+   | yes        |
 
 Ubuntu 24.04 LTS is not supported: it ships Podman 4.9.3 and will keep shipping
-that for its whole supported life. On any row marked "no", the engine has to
-come from somewhere other than the distribution:
+that for its whole supported life. On any row marked "no", `apt install podup`
+refuses rather than installing a podup that cannot reach an engine, and the
+engine has to come from somewhere other than the distribution:
 <https://podman.io/docs/installation>.
+
+Driving a **remote** Podman, or a `podman machine`, is the case apt cannot
+express: a package relationship only sees the local machine. Use the release
+binary there. `install.sh` warns instead of refusing when no local Podman is
+present, and takes `--skip-podman-check` when a local one is present but is not
+the engine podup will use.
 
 ### Platforms
 
--- a/debian/control
+++ b/debian/control
@@ -23,8 +23,26 @@ Rules-Requires-Root: no
 
 Package: podup
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}
-Recommends: podman (>= 5.0)
+# podman is a Depends, not a Recommends. podup is a translator: it turns a
+# compose file into calls on Podman's libpod API and does nothing else, so a
+# podup without an engine is not a degraded install, it is an install that
+# cannot do the one thing it is for.
+#
+# The floor is the same one MIN_LIBPOD_API_MAJOR enforces at run time and
+# PODMAN_MIN_MAJOR checks in install.sh. Three copies of one number, tied by
+# tests/podman_floor.rs, because as a Depends this one decides whether the
+# package installs at all rather than merely what apt suggests alongside it.
+#
+# The consequence, stated rather than discovered: on a distribution whose
+# podman is below the floor, `apt install podup` now FAILS instead of leaving
+# a podup that cannot reach an engine. Ubuntu 24.04 LTS ships 4.9.3 and is
+# already documented as unsupported, so apt now refuses what the README
+# already said would not work. The cost is the operator driving a REMOTE
+# socket or a `podman machine`, who has no local engine to satisfy this and
+# must now install one; install.sh serves that case with
+# --skip-podman-check, and apt has no equivalent because a package
+# relationship cannot ask where the socket is.
+Depends: ${shlibs:Depends}, ${misc:Depends}, podman (>= 5.0)
 Description: docker-compose translator and runner for rootless Podman
  podup parses docker-compose.yml files and drives rootless Podman to
  create the equivalent containers, networks and volumes. It provides
--- a/docs/debian-packaging.md
+++ b/docs/debian-packaging.md
@@ -97,7 +97,7 @@ is published and `apt upgrade` installs it — nothing for users to re-run.
 
 ## What the skeleton covers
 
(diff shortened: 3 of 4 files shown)


Jaro-c:
Two checks went red, `rust / Test (windows-latest)` and the `rust / Extra platforms` aggregate over it. Both were mine, and the fix is `69a83d8`.

`the_package_requires_podman_rather_than_suggesting_it` located the binary stanza with `split_once("\nPackage: podup\n")`. That is an exact byte match, so on a checkout with CRLF endings the newline after `podup` is `\r\n`, the needle is never found, and the `expect` panics. Windows runners check out with `core.autocrlf=true` and this repository has no `.gitattributes` forcing LF, so it failed there and nowhere else, over a file this change had not made platform-specific.

**Reproduced before edited.** Converting the working copy of `debian/control` to CRLF makes the real test panic at the real line with the real message, on Linux:

```
thread 'the_package_requires_podman_rather_than_suggesting_it' panicked at tests/podman_floor.rs:83:10:
debian/control has a binary stanza
```

That is what identified the mechanism. Reading the job log was not possible yet, because the run was still in progress and `gh run view --log` refuses until it completes.

The stanza is found with `lines()` now, which strips the carriage return, and it asserts it found relationship lines at all, so a parser that stops seeing the file reports that rather than reporting no violations.

**The class matters more than the instance, so it is a test.** The three floor parsers take a `&str` instead of reading the file themselves, and a new case feeds all of them the same files with CRLF endings and requires the same answer. Reverting the stanza parser to the byte search turns it red in a second on Linux, where the defect previously needed a Windows runner and a full round trip. The matrix leg that says "your parser assumes LF" is the one that costs the most to hear from.

Jaro-c:
Windows went red a second time, and the second failure was in the regression test I added for the first one. Same class, one level up. Fixed in `1161331`.

On a Windows checkout the file already holds CRLF, so building the fixture with `replace('\n', "\r\n")` produces `\r\r\n`, which is a shape no checkout delivers. `str::lines` strips one carriage return and leaves one on the line content, and `floor_from_source` ran `trim_end_matches(';')` **before** `trim()`: with a trailing `\r` the semicolon is not the last character, so it is never removed and `"5;"` fails to parse.

```
LF                            rest='5;'    -> parse('5')   ok
CRLF (Windows checkout)       rest='5;'    -> parse('5')   ok
CRLF doubled (the fixture)    rest='5;\r'  -> parse('5;')  panic
```

Two changes, because either alone leaves something wrong. The fixture normalises to LF before converting, so it asserts over a shape that exists. The parser trims whitespace before the semicolon rather than after, so the order stops deciding the answer.

**Verified in the failing condition, not around it.** The three files converted to CRLF on disk, which is what the Windows runner checks out, reproduce the panic at `podman_floor.rs:36` before the change and pass after it. Restored to LF, all four pass. Reintroducing the byte-search stanza parser still turns the regression test red, so the case it exists for is intact.

I also checked the red belonged to the new run rather than being the old one still displayed: the failing check was on `69a83d83`, the head at the time, not on the commit before it.

**Not done here, and worth its own change.** A `.gitattributes` with `* text=auto eol=lf` would remove this class from the repository rather than from these three parsers: no checkout would deliver CRLF and no future file reader would have to think about it. It touches every file's checkout behaviour for anyone on Windows, so it is not something to slip into a packaging change.

PR #1666, opened 2026-09-03
feat: x-podman-pod, one shared network namespace per project

### LINKED ISSUE #1665: Pods: one shared network namespace per project, opt-in with x-podman-pod
podup creates one container per service on a project network, the way docker compose does. Podman has a second shape for the same thing: a **pod**, one shared network namespace with an infra container, joined by every container of the project. Measured on 2026-09-03 with 42 `alpine` containers, three runs each: creation on a bridge network 4.65 to 4.72 s, inside a pod 3.27 to 3.60 s, about 33 ms less per container (the per-container network setup that netavark does once per pod); teardown equal. Inside a pod the services reach each other on `localhost`, the whole project is one network namespace to audit, and `podman pod` becomes a unit of operation (`stats`, `logs`, `kube generate`).

It changes semantics, so it is opt-in and rejects what does not fit.

## What I want

A top-level extension, `x-podman-pod: true`, that puts every service of the project into one pod named after the project.

- **The pod.** `up` creates the pod through `POST /libpod/pods/create` when it does not exist, before the containers, with: `name` = the project name, labels `podup.project=<project>` and `podup.pod-config-hash=<hash>`, `shared_namespaces: ["net"]` only (UTS and IPC stay per container, so `hostname:` keeps working), `networks` = every non-external network the file declares plus the external ones by name, `portmappings` = the union of every service's `ports:` (a pod publishes ports, a container inside it cannot), and `hostadd` = one `<service>:127.0.0.1` entry per service, so `db:5432` written in a compose file resolves to the shared namespace the way it resolves on a compose network. The hash covers what only a pod recreate can change: the port set, the network set and the host entries. When the pod exists and its hash differs, `up` recreates it (containers down, pod removed, pod created, containers up) and says so with the `Recreating` vocabulary on a line for the pod.
- **The containers.** The `SpecGenerator` gets `pod: <project>`, no `portmappings`, no `networks`, no `netns`; everything else is untouched. `depends_on`, healthchecks, `scale`, secrets, the config hash and the image-ID comparison work as before; a replica is one more container in the pod.
- **What is refused,** at validation time before anything is created, each with a message naming the service and the key: `network_mode` on any service; a service whose `networks:` set differs from another service's (all services must be on the same networks, or declare none and get the project default); two services publishing the same host port (today two containers on a network would also collide, but the pod reports it before Podman does); `ports:` with a host IP that differs between services for the same port. Two services listening on the same container port inside the pod will conflict at runtime and podup cannot see it; the doc says so.
- **`down`** removes the containers as today, then `DELETE /libpod/pods/<project>?force=true` (that removes the infra container), then networks and volumes as today. `down` on a project that was started without the extension is unchanged. A pod left behind by a crashed `up` is removed by `down` and by `down --remove-orphans` through the `podup.project` label.
- **`ps`** shows the containers as today; the infra container is not listed.
- **`generate quadlet`** emits one `<project>.pod` unit (`[Pod]` with `PodName=`, one `Network=` per declared network, one `PublishPort=` per port, one `AddHost=` per service) and each `.container` gets `Pod=<stem>.pod` and loses its own `PublishPort=` and `Network=` lines. Quadlet supports `.pod` units since Podman 5.0, which is the floor.
- **Docs:** the extension in `docs/commands.md`'s table and a section on what changes and what is refused; a row in `docs/threat-model.md` under "The engine boundary" (one namespace, one audit surface); the benchmark scenario `wide-level` gains a pod variant only if the numbers are published, which is a separate change.

## Acceptance

- Unit tests against the fake Podman: the pod create body (name, labels, shared namespaces, ports as the union, networks, host entries); containers created with `pod` set and no ports or networks of their own; the hash changing when a port is added and not when a service's command changes; each refusal with its message; `down` deleting the pod after the containers; the Quadlet `.pod` unit and the `Pod=` line with the ports moved.
- Integration on both Podman majors: `up` with two services and the extension, `podman pod ps` shows one pod with two containers plus infra; from the first container `getent hosts <second service>` resolves to `127.0.0.1` and a TCP connect to the second service's port on that name succeeds; a port change recreates the pod with `Recreating` printed for it; `down` leaves no pod.
- Every new test goes red when its control is deleted; `cargo mutants` before merge.

Jaro-c:
Fixes #1665.

## What

`x-podman-pod: true` at the top level puts every service of the project into one Podman pod named after the project: only the network namespace shared, ports published by the pod as the union of every service's list, the project's networks, and a `<service>:127.0.0.1` host entry per service so `db:5432` keeps resolving. A hash of ports, networks and host entries is a label on the pod; a later `up` that computes a different one recreates the pod and creates every container afresh. `down` removes the pod after the containers. `ps` hides the infra container. `generate quadlet` writes a `<project>.pod` unit and each `.container` references it. `network_mode`, divergent `networks:` sets and duplicate host ports are refused before anything is created. Docs: a Pods section in `docs/commands.md`, a row in the threat model.

## How it was built

MiniMax wrote steps 1 to 3 (types, refusals, engine) with 11 unit tests and stopped when its plan ran out of tokens; I finished the Quadlet unit, the docs and the integration tests, and reviewed the rest.

What the review and the real engine found, each with a test now:

- `up` fetched the container list before ensuring the pod, so a pod recreate (which removes the members) left `up` reading gone containers as unchanged and only starting them. `ensure_pod` now reports a recreate and `up` forgets the list; the test fails without that line.
- libpod refuses a pod that names networks without `netns: bridge`; the fake did not know. Added, asserted on the create body.
- The infra container carries the project label and no service label, so it was reported, and would have been removed, as an orphan. Excluded by `IsInfra`.
- The `.pod` unit references declared networks through their `.network` units, like `.container` units do.

## Verified

- Unit: 1806 pass. Integration on Podman 5.7.0: one pod with both services and `SharedNamespaces: [net]`; `getent hosts db` inside `web` gives `127.0.0.1` and a TCP connect to `db:5432` by name succeeds; a port change recreates the pod (new ID, `Recreated` printed, no orphan warning); `down` leaves no pod.
- `cargo mutants` over the pod module runs next; the result lands in this PR before merge.


Jaro-c:
Measured with podup itself (not the CLI) on the `wide-level` scenario, 42 services, 10 runs after 2 warm-up, twice, same release build of this branch, cores 2-9:

| | project network | pod |
|---|---|---|
| `up -d` | 1.137 / 1.133 s | 1.499 / 1.498 s |
| `down -v` | 1.907 / 1.764 s | 1.390 / 1.401 s |

A pod is slower to create under podup's concurrent `up` and faster to tear down; the CLI measurement in the issue (one create at a time) went the other way. The Pods section in `docs/commands.md` now says so and names the reasons to pick a pod: `localhost` between services, one namespace, one place ports are published.

Mutation sweeps: 34 of 38 pod mutants caught on the first pass; the survivors got tests, and the host-IP branch two of them sat in turned out unreachable behind the duplicate-port rule and was removed. Full integration suite on this branch against Podman 5.7.0: 196 of 196.

PR #1693, opened 2026-09-04
fix(ci): the macOS gate reads otool's value line, and the arm64 Windows leg signs with an x64 Python

### LINKED ISSUE #1692: The v5.9.0 release run failed on the macOS gate and on signing the arm64 Windows asset
Run 33911807587 on tag v5.9.0, 2026-09-04. Nothing was signed or published: the gates stop the leg before signing.

**Both darwin legs:** `FAIL not-macho podup-darwin-arm64` and the same for x86_64. `check-hardening-macos.sh` read the flags with `awk '/^flags/'`, but `otool -h` prints `flags` as the last column of its header line and the value on the next line, which starts with the magic. The fixture that would have caught it skips where `otool` is absent, and no job ran it on a Mac.

**windows-arm64:** the gate passed; `Sign binary` failed because `cryptography==50.0.1` publishes no `win_arm64` wheel, pip tried to build it from source under `--require-hashes` and the build needs OpenSSL. The comment above the matrix entry said exactly this, and #1683 moved the leg to the arm64 runner past it.

Acceptance: the script reads the value line; a macOS job runs the fixture on a Mac and reads the properties off a binary built like the release; the arm64 leg signs with an x64 Python the runner emulates, and the pull-request job on `windows-11-arm` installs the signing dependencies the same way so a wheel that stops resolving fails before a tag; the release is re-run on the same tag by dispatch.

Jaro-c:
Closes #1692. Against `main`, since the release runs from the tag and is re-run by dispatch once this lands; `develop` gets it by the usual sync.

- `check-hardening-macos.sh` reads the flags off `otool -h`'s value line (the one starting with the magic), not the column-name line.
- `lint-shell.yml` gains a `macos-latest` job: the fixture with its clang-built controls, then a release-profile build and the gate on the real binary. That job runs on this pull request, which is the proof the script was missing.
- The windows-arm64 leg sets up an x64 Python 3.12 before signing; the sign step uses `python` when a setup step provided it. The `windows-11-arm` pull-request job installs the signing dependencies under `--require-hashes` the same way, and also runs here.
- The matrix comment that said the wheels were x64-only, which #1683 moved past, now says what the leg does about it. Threat-model row updated.


Jaro-c:
Closing: pull requests into main must come from develop (the main guard is right), and the script fix has to be in the tag anyway, since a dispatched release checks out the tag. Reopened against develop as the next pull request.

Issue #1735, opened 2026-09-07
autostart install --mode quadlet skips the quadlet validator


Jaro-c:
`internal/generate.rs:82` calls `validate_for_quadlet(file)?` before rendering. `internal/autostart/quadlet.rs:137` calls `quadlet::generate_at` with no validation at all, and `validate_for_quadlet` is private to the bin crate, so the library path cannot reach it even if it wanted to.

The path that skips the validator is the one that writes the units and boots them.

## Reproduction

Same `compose.yaml`, carrying `mem_limit: "512m --privileged"`:

```
$ podup generate quadlet
podup: error: unsupported feature: service 'web': mem_limit '512m --privileged'
       is not a valid memory size

$ podup autostart install --mode quadlet --dry-run
17:PodmanArgs=--memory=512m --privileged
autostart exit=0
```

One command refuses the file. The other emits the flag and exits 0.

## Why this is its own issue and not part of #1734

#1734 is the escaping defect: values reach `PodmanArgs=` unquoted. This is the containment defect: the validator that catches the malformed value exists, works, and is not on the path that matters. Fixing either one alone leaves the other standing, and the fix for this one is a move rather than a rewrite.

## Acceptance

- `validate_for_quadlet` lives in the library, and both `generate quadlet` and `autostart install --mode quadlet` call it.
- A test that runs the same file through both commands and requires the same verdict. The current pair of outputs above is the fixture.
- The test names the property, not the call: it must fail if a third quadlet-writing path is added later without validation.

Jaro-c:
Landed in #1749, squashed into `develop`. Closed by hand: `Closes #N` only fires on merges into the default branch.

PR #1749, opened 2026-09-07
fix(quadlet): quote the values we interpolate into PodmanArgs=

### LINKED ISSUE #1734: Seven compose fields smuggle arbitrary podman run flags through PodmanArgs=
`internal/quadlet/render.rs:242-247` exempts three keys from escaping:

```rust
/// Keys whose value is an argument line we render (and quote) ourselves
/// (`PodmanArgs`, `Exec`, `Entrypoint`); systemd must keep splitting these on
/// whitespace, so they are passed through with only control characters stripped.
fn key_is_arg_line(key: &str) -> bool {
	matches!(key, "PodmanArgs" | "Exec" | "Entrypoint")
}
```

"we render (and quote) ourselves" is true for `Exec` and `Entrypoint`, which go through `render_command` and `quote_exec_arg`. It is false for `PodmanArgs`. Seven sites `format!` raw compose strings into it, and `escape_unit_value` then returns early at `render.rs:268-270`, skipping the `%` doubling and the quoting every other key gets:

```
internal/quadlet/unit/security.rs:105   --security-opt apparmor={profile}
internal/quadlet/unit/container.rs:286  --memory={mem}            (mem_limit)
internal/quadlet/unit/container.rs:298  --cpus={c}
internal/quadlet/unit/container.rs:301  --cpuset-cpus={cs}
internal/quadlet/unit/container.rs:453  --memory={mem}            (deploy.resources.limits.memory)
internal/quadlet/unit/build.rs:115      --build-arg {key}={v}
internal/quadlet/unit/build.rs:116      --build-arg {key}
```

## Reproduction

```yaml
name: victim
services:
  web:
    image: alpine:3.20
    security_opt: ["apparmor=unconfined --privileged --net=host -v /:/hostfs"]
    cpuset: "0 --privileged -v /:/hostfs2"
```

`generate quadlet` exits 0 with no warning and writes:

```
PodmanArgs=--cpuset-cpus=0 --privileged -v /:/hostfs2
PodmanArgs=--security-opt apparmor=unconfined --privileged --net=host -v /:/hostfs
```

Fed to Podman 5.7.0's own generator:

```
$ QUADLET_UNIT_DIRS=out /usr/lib/systemd/system-generators/podman-system-generator -dryrun -user
ExecStart=/usr/bin/podman run --name victim-web ... --cpuset-cpus=0 --privileged \
  -v /:/hostfs2 --security-opt apparmor=unconfined --privileged --net=host \
  -v /:/hostfs alpine:3.20
```

Host root bind-mounted into a privileged container on the host network. I reproduced the `PodmanArgs=` line myself from a fresh `cargo build` of `develop`.

## Both controls built to surface this go silent

Two files differing only in how `--privileged` is spelled, same host, same binary:

| | `privileged: true` | smuggled through `security_opt` |
|---|---|---|
| `container/host_mode.rs:65` warning | fires | silent |
| `podup audit --format json` | includes `{"check":"privileged"}` | `privileged` absent, the other six findings byte-identical |
| resulting `podman` argv | `--privileged` | `--privileged` |

`%` is also not doubled on this path, so an injected `-v %h:/homedir` is expanded by systemd into the operator's home. `Environment=` correctly emits `%%h`.

## Scope, established by testing rather than assumed

Hostile whitespace pushed through 17 other string fields comes out as one quoted argv element: `Image`, `PublishPort`, `ShmSize`, `HostName`, `WorkingDir`, `User`, `DNS`, `Tmpfs`, `Sysctl`, `Volume`, `HealthCmd`, `Pull`, `AddHost`, and `AddDevice`/`AddCapability`/`Mask` split but Quadlet re-prefixes each token with its own flag. Newline injection is fully blocked: a service key of `"web\nExecStartPre=/bin/sh -c id\n"` produced a sanitised one-line unit and could not forge the `# podup-owner:` marker. The defect is this key and only this key.

## It contradicts two documents we ship

`docs/threat-model.md:65` says Quadlet values are escaped. `docs/security-model.md:236` says they are filtered through `escape_unit_value`. So this is not the declared residual at `threat-model.md:132`: that residual says a hostile compose file can request any privilege **visibly**, and this makes the request invisible to the two controls built to surface it.

## Acceptance

- The arg-line keys podup builds itself quote their interpolated value, so `--cpuset-cpus="0 --privileged -v /:/hostfs"` is one argv element that podman rejects.
- A test per injection site, each asserting the resulting argv rather than the unit text, since the unit text is what looked correct.
- A test that the `privileged` audit finding and the `host_mode.rs:65` warning both fire on the smuggled spelling, because their silence is the reason this is worth more than a rendering bug.
- `%` doubling applies on this path too, with a case for `%h`.
- The two documents above are corrected or the code is made to match them. Whichever way it goes, they must not keep claiming a property the tree does not have.

Jaro-c:
`escape_unit_value` exempted `PodmanArgs=` on the grounds that we render
and quote that line ourselves. True for `Exec` and `Entrypoint`, false
for `PodmanArgs`: seven sites interpolated raw compose strings into it,
so one `security_opt` entry put `--privileged --net=host -v /:/hostfs`
into the argv Podman's own generator builds. Host root, bind-mounted
into a privileged container on the host network, from a compose file.

`quote_podman_arg_value` now wraps each interpolated value at those
seven sites, strips control characters and doubles `%`, which that path
was also missing: an injected `-v %h:/homedir` was expanded by systemd
into the operator's home. `Environment=` already did this and is what
the helper follows.

Measured against Podman 5.7.0's generator on the file from the issue,
after the change:

    "--cpuset-cpus=0\x20--privileged\x20-v\x20/:/hostfs2"
    --security-opt "apparmor=unconfined\x20--privileged\x20--net=host\x20-v\x20/:/hostfs"

One argv element each. No loose `--privileged`.

The scope stays exactly this key. Seventeen other string fields already
came out as a single quoted element and are untouched.

The tests assert the argv Podman would build rather than the unit text,
because the unit text is what looked correct for the whole life of this
bug. Making `quote_podman_arg_value` return its input unchanged turns
eight of them red, including the two named for the smuggling itself, and
the tree is green with it restored.

`docs/security-model.md` claimed Quadlet values are filtered through
`escape_unit_value`. That sentence was false for this key and now names
both helpers.

One acceptance criterion I wrote in the issue was wrong and is not
implemented: it asked for the `privileged` audit finding to fire on the
smuggled spelling. After this change the smuggled spelling grants no
privilege, so reporting one would be a false positive. Measured: the
honest `privileged: true` file still reports `privileged`, and the
smuggled file reports the same six findings without it, which is now the
correct answer rather than the silence the issue complained about.

Closes #1734.

### DIFF
--- a/docs/security-model.md
+++ b/docs/security-model.md
@@ -233,9 +233,9 @@ data into a process or onto disk:
   a filesystem path from it.
 - **URL paths** go through `urlencoded` so container names, project names,
   and tags with arbitrary bytes reach libpod as a single encoded segment.
-- **Quadlet values** are filtered through `escape_unit_value` /
-  `safe_unit_stem`; the unit filename is checked again by `write_units` so
-  a poisoned value cannot break out of the output directory.
+- **Quadlet values** are filtered through `escape_unit_value` and the dedicated
+  `PodmanArgs=` interpolations use `quote_podman_arg_value`; the unit filename is
+  checked again by `write_units` so a poisoned value cannot break out of the output directory.
 - **Signal names** are resolved to numbers through `resolve_stop_signal`:
   the libpod endpoint is a number, a string returns HTTP 500.
 - **Pull policies** are normalised through `libpod_pull_policy`; an
--- a/internal/autostart/quadlet.rs
+++ b/internal/autostart/quadlet.rs
@@ -134,6 +134,7 @@ pub fn install_quadlet<S: SystemCtl>(
 		)));
 	}
 
+	quadlet::validate_for_quadlet(file)?;
 	let result = quadlet::generate_at(file, project, base_dir);
 	if let Some(dup) = result.duplicate_filename() {
 		return Err(ComposeError::Autostart(format!(
--- a/internal/generate.rs
+++ b/internal/generate.rs
@@ -4,8 +4,6 @@
 use std::io::Write;
 use std::path::Path;
 
-use podup::compose::types::ComposeFile;
-
 /// Quadlet units are systemd unit files; they only run on Linux hosts (where
 /// systemd consumes them from `~/.config/containers/systemd/`). Generating them
 /// on macOS/Windows is legitimate (e.g. to deploy to a remote Linux host), so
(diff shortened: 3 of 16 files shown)


Jaro-c:
This also closes #1735. `validate_for_quadlet` moved into `podup::quadlet` and both callers use it, so the same file now gets the same verdict from either command:

```
$ podup -f mem.yaml generate quadlet
podup: error: unsupported feature: service 'web': mem_limit '512m --privileged' is not a valid memory size
$ podup -f mem.yaml autostart install --mode quadlet --dry-run
podup: error: unsupported feature: service 'web': mem_limit '512m --privileged' is not a valid memory size
```

Before this change the second one printed `PodmanArgs=--memory=512m --privileged` and exited 0.

PR #1769, opened 2026-09-08
fix(ui): one tail entry per line, so the board erases what it painted

### LINKED ISSUE #1733: [bug]: a builder stream chunk with embedded newlines is one note, so the board never erases its last frame
### Component

cli / ui (`podup build` live board) — `internal/ui/progress/mod.rs`, `internal/ui/progress/row.rs`

### What happened?

`podup build --no-cache` repainted the live board roughly eighty times without erasing the previous frame, so one build scrolled the same block of builder output down the terminal over and over. The build itself was fine; only the drawing was wrong.

Each frame looked like this, with the elapsed time ticking up and everything else identical:

```
[+] Running 0/1
 ⠙ Image     alanalarana:latest Building   2.0s
 Attempting to install security scanner from npm...
 Security scanner installed successfully.
 ⚠ Socket Security Scanner free mode. Set SOCKET_API_KEY to use your Socket org settings.
 + @biomejs/biome@2.5.12
+ @tailwindcss/postcss@4.3.3
+ @types/node@26.4.1
+ @types/react@19.2.18
+ @types/react-dom@19.2.7
+ babel-plugin-react-compiler@1.0.0
+ postcss@8.5.28
```

**The tell is the indentation.** `render_note` prefixes each note with exactly one space. The first `+ @biomejs/biome@2.5.12` has that space; the six lines under it do not. All seven are one note.

That is a single `VecDeque` entry holding embedded newlines. The board counts it as one line, the terminal draws it as seven, and from that point on `painted` is short by six.

`Region`'s own doc comment predicts this before it happens:

> Owns the count of rows it last painted, which is the only state the repaint arithmetic needs, and the reason every line handed to it must already be truncated to the terminal width. A line that wraps makes the terminal count two rows where this counted one, and from then on every repaint erases the wrong lines.

The invariant is right. It just does not hold for a note, and wrapping is not the only way to break it — an embedded `\n` breaks it the same way and is not covered by the width fit.

The repetition stops on its own once the block stops changing, which is consistent: `cursor_up(painted)` lands in the wrong place every frame, but once the content is identical the wrong place keeps redrawing the same bytes and the damage stops being visible.

**Expected:** one board, repainted in place, whatever the builder emits.

### Steps to reproduce

Any build whose builder writes more than one line per stream chunk. Mine is a Bun install inside a multi-stage `Containerfile`; `bun install` prints its added-package list as a block.

```
podup -f compose.yaml build --no-cache
```

A smaller reproducer is a `RUN printf 'a\nb\nc\nd\ne\n'` early in a Containerfile: any single write carrying more than one newline should do it.

### Where it is

`internal/ui/progress/mod.rs`, in `note_for`:

```rust
let trimmed = line.trim();
if trimmed.is_empty() {
    return;
}
```

`trim()` removes leading and trailing whitespace and leaves interior newlines alone, so `trimmed` can be a multi-line string. It then goes to `push_note_live` whole:

```rust
let tail = notes.entry((kind, name.to_string())).or_default();
tail.push_back(trimmed.to_string());
```

and comes back out as one element in the repaint loop:

```rust
if let Some(tail) = notes.get(&(r.kind, r.name.clone())) {
    for line in tail.iter() {
        lines.push(row::render_note(line, width));
    }
}
```

`render_note`'s truncation does not save it either, because `chars().take(width)` counts the `\n` characters as ordinary chars: a seven-line chunk of 200 characters is under an 80-column `width` only in the sense that it is not, and either way what comes out still contains newlines.

There is a second consequence worth noting: `MAX_NOTES_PER_ROW` is meant to cap the tail at a few lines, and a multi-line note defeats it. One entry can be arbitrarily tall, so the tail can grow past its own limit.

### Suggested fix

Split at ingestion, so one entry is always one terminal row. `note_for` is the right place, since it already owns the trim-and-drop-empty step:

```rust
// One entry per line. A stream chunk can carry several lines in one write,
// and the repaint arithmetic counts entries: a note holding `\n` is drawn as
// several rows but counted as one, and every later repaint then erases too
// few. It also lets a single entry defeat MAX_NOTES_PER_ROW.
for line in line.split('\n') {
    let trimmed = line.trim();
    if trimmed.is_empty() {
        continue;
    }
    // ... existing per-line body
}
```

`split('\n')` rather than `lines()` on purpose: `lines()` also swallows a trailing `\r`, and the following `trim()` handles `\r\n` anyway, so `split` keeps the two responsibilities separate.

Worth pairing it with a defensive guard in `render_note`, so the invariant is enforced where it is relied on rather than only where it happens to be produced today:

```rust
debug_assert!(!line.contains('\n'), "a note must be one line: {line:?}");
```

Tests I would add:

- `note_for` handed `"a\nb\nc"` produces three entries, not one.
- A note carrying `\r\n` produces clean entries with no stray `\r`.
- `MAX_NOTES_PER_ROW` still holds when the input arrives as one multi-line chunk.
- Empty lines inside a chunk are dropped, matching the current single-line behaviour.

I have not checked whether the same path feeds `pull`'s layer counting; if it shares `note_for`, it has the same exposure whenever a registry sends two progress lines in one chunk.

Happy to open the PR.

### podup version

v5.9.1

### How did you install podup?

apt

### Podman version and major

5.4.2 (Podman 5)

### Podman socket

unix:///run/user/1002/podman/podman.sock

### Operating system

Debian 13 trixie

### Architecture

x86_64

### Is this a rootless install?

Yes, rootless

### Relevant logs

```shell
==> Building every layer from scratch
[+] Running 0/1
 ⠙ Image     alanalarana:latest Building   2.0s
 Attempting to install security scanner from npm...
 Security scanner installed successfully.
 ⚠ Socket Security Scanner free mode. Set SOCKET_API_KEY to use your Socket org settings.
 + @biomejs/biome@2.5.12
+ @tailwindcss/postcss@4.3.3
+ @types/node@26.4.1
+ @types/react@19.2.18
+ @types/react-dom@19.2.7
+ babel-plugin-react-compiler@1.0.0
+ postcss@8.5.28
[+] Running 0/1
 ⠹ Image     alanalarana:latest Building   2.0s
 Attempting to install security scanner from npm...
 ... (the same block, ~80 times, only the elapsed time changing)
 ✔ Image     alanalarana:latest Built  28.8s
```

Jaro-c:
A builder stream chunk carrying several lines in one write became a
single tail entry. The repaint arithmetic counts entries, so the board
counted one row where the terminal drew seven, `cursor_up(painted)`
landed six lines short from that frame on, and one build scrolled the
same block down the terminal about eighty times.

`Region`'s doc comment already stated the invariant: a line that wraps
makes the terminal count two rows where this counted one, and from then
on every repaint erases the wrong lines. The invariant was right and an
embedded newline breaks it by a route the width fit does not cover, since
`chars().take(width)` counts `\n` as an ordinary character.

`note_lines` splits at ingestion, which is also where the trim and the
drop-empty already lived, so one entry is one row again. `split('\n')`
rather than `lines()` on purpose: `lines()` also swallows a trailing
`\r`, and the `trim()` handles `\r\n` anyway, so the split and the
trimming stay two decisions.

It restores `MAX_NOTES_PER_ROW` as well. The cap counts entries, so one
arbitrarily tall entry defeated it.

`render_note` gains a `debug_assert!` that a note carries no newline, so
the invariant is enforced where it is relied on rather than only where it
is produced today.

The split is a pure function with its own test hook, because the live
path needs a region the cargo-test runner cannot give us and the
existing hook enters below the split. Replacing it with a
`std::iter::once` turns the multi-line case red on its own.

Closes #1733.

### DIFF
--- a/internal/ui/progress/mod.rs
+++ b/internal/ui/progress/mod.rs
@@ -227,6 +227,31 @@ pub fn start_anchored(
 	emit(sink, kind, name, verb);
 }
 
+/// The lines one stream chunk contributes to a row's tail.
+///
+/// A builder writes several lines in one chunk, and the repaint arithmetic
+/// counts entries: a note holding `\n` is drawn as several terminal rows and
+/// counted as one, so `cursor_up(painted)` lands short and every later
+/// repaint erases too few lines (#1733). Splitting here keeps the invariant
+/// `Region` documents, that one entry is one row.
+///
+/// `split('\n')` rather than `lines()` on purpose: `lines()` also swallows a
+/// trailing `\r`, and the `trim()` here handles `\r\n` anyway, so the two
+/// responsibilities stay apart. Empty parts are dropped, matching what a
+/// single blank line already did.
+fn note_lines(line: &str) -> impl Iterator<Item = &str> {
+	line.split('\n')
+		.map(str::trim)
+		.filter(|part| !part.is_empty())
+}
+
+/// Test hook for [`note_lines`], which is otherwise only reachable through a
+/// live region the cargo-test runner cannot give us.
+#[cfg(test)]
+pub(crate) fn note_lines_for_tests(line: &str) -> Vec<&str> {
+	note_lines(line).collect()
+}
+
 /// Hand one line of a row's stream output to the renderer.
 ///
 /// On a live terminal the line is appended to the per-row tail kept under the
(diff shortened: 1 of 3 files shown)


Jaro-c:
Measured end to end, not only at the unit level. A real `podup build --no-cache` under a pseudo-TTY, control at `develop` against this branch, with a `RUN printf` writing five lines in one chunk.

The metric is the gap between the rows a frame actually paints and the rows `cursor_up` claims to erase:

| offset (painted minus declared) | control | this branch |
| --- | --- | --- |
| 0, correct | 8 frames | **45 frames** |
| +4, the defect | **33 frames** | 0 |
| +1 | 2 | 4 |
| -5, the closing frame | 1 | 1 |

The +4 is the five-line note counted as one, and it is gone.

The remaining +1 frames are the board legitimately growing: `cursor_up` erases what the *previous* frame painted, so a frame that adds a row is one short by construction. Checked on each: `declared = painted - 1` every time.

The control also reproduces the indentation tell from the issue, ` linea-a` with the leading space `render_note` adds and `linea-b` onward without it, which is what says all five were one entry.

PR #1776, opened 2026-09-10
fix(engine): pack tars without GNU sparse entries, which Podman refuses

### LINKED ISSUE #1775: A sparse file in the build context makes podup emit a GNU sparse entry that Podman rejects
## What happens

A build whose context contains a sparse file fails against the libpod endpoint:

```
podup: error: podman error: podman API error (HTTP 500): unhandled tar header type 83
```

`podman build` on the identical context succeeds, so the report reads as "podup cannot build what Podman can build". That is exactly what it is.

## Why

`tar::Builder::new` in the `tar` crate defaults `sparse` to true:

```
~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tar-0.4.46/src/builder.rs:41
                sparse: true,
```

With that on, `append_file` detects holes with `SEEK_HOLE`/`SEEK_DATA` and writes a GNU sparse entry instead of a plain one. The GNU sparse typeflag is `b'S'`, which is decimal 83:

```
~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tar-0.4.46/src/entry_type.rs:62
            b'S' => EntryType::GNUSparse,
```

The server side is `containers/storage/pkg/archive`, whose `createTarFile` has no case for that type and returns `unhandled tar header type %d`. So podup is not corrupting the stream, it is deliberately writing an entry type the receiver never accepts.

We construct a `tar::Builder` in four places and not one of them turns the default off:

- `internal/engine/build/context.rs:104`
- `internal/engine/build/context.rs:143`
- `internal/engine/watch/sync.rs:28`
- `internal/engine/copy/archive.rs:22`

Podman's own client packs the context in Go and never emits `'S'`, which is why the CLI is unaffected.

## Reproduction, measured on podup v5.9.2 with podman 5.7.0

```sh
mkdir -p ctx && cd ctx
printf 'FROM docker.io/library/alpine:3.20\nCOPY . /ctx\nRUN ls -l /ctx\n' > Containerfile
dd if=/dev/zero of=big.sparse bs=1 count=0 seek=1048576 status=none
printf 'data' >> big.sparse
```

with a compose file whose single service builds that context.

| Case | Result |
| --- | --- |
| context contains `big.sparse` | `podman API error (HTTP 500): unhandled tar header type 83`, exit 1 |
| same file rewritten with `cp --sparse=never` | build succeeds, exit 0 |
| `podman build` on the sparse context | build succeeds, exit 0 |

The second row is the control that pins the cause to the hole in the file and not to anything else in the context.

## How people hit this without knowing it

Nobody deliberately puts a sparse file in a build context. They arrive on their own: database files, disk images, log files preallocated with `fallocate`, virtual machine state, anything copied out of a `podman machine` directory. The user does not see a difference in `ls`, so the failure looks arbitrary. The first report I got was from Windows, where the sparse file came from the WSL side, and the reporter concluded podup's build was broken in general.

## Fix

Call `.sparse(false)` on every builder we construct, so we emit plain entries the way Podman's own client does. That is the behavior every consumer of these tars expects: the build endpoint, the `cp` endpoint and the watch sync path all feed the same server-side archive reader.

## Acceptance

- A test packs a context containing a real sparse file and asserts that no entry in the resulting archive has entry type `GNUSparse`. Asserting the header bytes, not asserting that the call returned `Ok`.
- The test sabotages cleanly: removing the `.sparse(false)` call makes it fail. A test that stays green with the fix reverted has not measured anything.
- The same coverage applies to the `cp` and watch sync builders, not only to the build context, since they write into the same server-side reader.

Jaro-c:
Closes #1775.

A file with holes on disk went out as a GNU sparse entry, typeflag `S`, decimal 83, because `tar::Builder::new` enables sparse mode by default. Podman has no case for that type, so it drops the whole archive on account of one file. The build endpoint says `unhandled tar header type 83`; the container archive endpoint that `cp` and the watch sync PUT to says `unrecognized Typeflag S`.

## What I changed

The four places that constructed a `tar::Builder` now go through one constructor, `engine::tar_stream::builder`, which turns the flag off. Turning it off four times would have fixed today and left the fifth site free to be written the old way, so `tests/tar_builder_single_site.rs` keeps it the only one. Those four sites sat in three unrelated files and were wrong the same way precisely because each was written on its own.

`follow_symlinks` is untouched at all three sites that set it, which matters: two of them set it to `false` on purpose so a symlink in the tree is stored as a link instead of packing the bytes of its target.

## Verification

Every row below is a real run on this machine, podman 5.7.0, released `v5.9.2` against this branch's binary, same input, and in every case the control is the same bytes with the hole filled in (`cp --sparse=never`). Fixtures were checked for a real hole (`stat -c '%s %b'`) before each run, on tmpfs and again on ext4.

| Path | Case | v5.9.2 | This branch |
| --- | --- | --- | --- |
| build | context holding an 8 MiB sparse file | `unhandled tar header type 83`, exit 1 | builds, exit 0 |
| cp | single file | `unrecognized Typeflag S`, exit 1 | copies, exit 0 |
| cp | directory holding a sparse file and a plain one | `unrecognized Typeflag S`, exit 1, and neither file arrives | both arrive |
| watch | initial sync | warning only, exit 0, `/app` empty | `synced`, both files present |
| watch | live event | warning only, exit 0, `/app` empty | `synced`, file present |

The `watch` rows are the reason this is worth a gate. A failed sync is a `warn!` and the loop is meant to survive one, so the exit code stays 0, the session keeps saying it is watching, and the file just never arrives. Verified by `ls` and `sha256sum` inside the container rather than by exit code.

No corruption from expanding the holes: every destination `sha256sum` equals the host's, and every byte count matches, on all five transfers.

Two things the runs turned up that were not in the issue. The rejection is per archive, so a sparse file takes its ordinary neighbours down with it. And `cp` of a directory creates the destination before the stream is refused, leaving an empty directory that reads as success to anything checking only for the path.

## Cost

Measured through the same `GzEncoder` the real paths use, on 1 GiB of holes with four bytes of data: 125 gzipped bytes become 1043015, and peak RSS moves from 2832 KiB to 2852. `io::copy` writes through a fixed stack buffer into the encoder, so the streaming build path still never holds the archive. About 1 MiB on the wire per GiB of holes, and the reading and gzipping of that GiB, which is the real cost.

## Tests

- `our_builder_never_writes_a_gnu_sparse_entry` packs a real sparse file through the constructor and asserts no entry is `GNUSparse`. Reverting `.sparse(false)` makes it fail; I checked, rather than assuming.
- `the_crate_default_emits_the_type_podman_refuses_when_the_hole_is_real` is the positive control. Without it the test above is vacuous wherever the crate never looks for holes, and would stay green with the fix reverted while reporting that it had checked something. The crate only looks on Android, FreeBSD and Linux, so the control is gated on exactly that list rather than on `cfg(unix)`: macOS is Unix and APFS really does leave the range unallocated, so a block-count check would have failed the control on the macOS runner for a reason unrelated to this fix.
- `the_expanded_entry_carries_every_byte_the_file_had` reads the archive back and checks the hole comes out as zeroes with the tail intact.
- `tests/tar_builder_single_site.rs` refuses a second construction site. It rejects both spellings: the qualified `tar::Builder::new(`, and an import of `tar::Builder` that would let a bare `Builder::new(` slip past a literal search. I sabotaged it with each spelling and watched it go red for each.

Full suite green: 2671 tests, 0 failures, plus `cargo fmt --check` and `cargo clippy --all-features --all-targets` with nothing emitted.

## Scope

podup only. The five other Rust repositories do not carry the `tar` crate even transitively, and there is no `.sparse(` call anywhere else in the tree.

### DIFF
--- a/internal/engine/build/context.rs
+++ b/internal/engine/build/context.rs
@@ -145,7 +145,7 @@ pub(super) fn stream_build_context<W: std::io::Write>(
 ) -> Result<()> {
 	let (ignore_name, ignore_patterns) = ignore_file(context);
 	let encoder = GzEncoder::new(writer, Compression::default());
-	let mut tar = tar::Builder::new(encoder);
+	let mut tar = crate::engine::tar_stream::builder(encoder);
 
 	// Force-include the active Dockerfile so an ignore file that matches it
 	// cannot drop it from the context the builder receives (Docker parity).
@@ -184,7 +184,7 @@ pub(super) fn stream_build_context_with_inline<W: std::io::Write>(
 ) -> Result<()> {
 	let (ignore_name, ignore_patterns) = ignore_file(context);
 	let encoder = GzEncoder::new(writer, Compression::default());
-	let mut tar = tar::Builder::new(encoder);
+	let mut tar = crate::engine::tar_stream::builder(encoder);
 
 	let mut header = tar::Header::new_gnu();
 	header.set_size(inline.len() as u64);
--- a/internal/engine/copy/archive.rs
+++ b/internal/engine/copy/archive.rs
@@ -19,7 +19,7 @@ pub(super) fn pack_path(
 	name_override: Option<&str>,
 ) -> Result<Vec<u8>> {
 	let encoder = GzEncoder::new(Vec::new(), Compression::default());
-	let mut tar = tar::Builder::new(encoder);
+	let mut tar = crate::engine::tar_stream::builder(encoder);
 	// `-L/--follow-link`: archive the symlink target's contents instead of the
 	// link itself.
 	tar.follow_symlinks(follow_link);
--- a/internal/engine/mod.rs
+++ b/internal/engine/mod.rs
@@ -52,6 +52,7 @@ mod staging;
 mod stats;
 pub use staging::is_safe_project_name;
 pub use stats::StatsOptions;
+mod tar_stream;
 mod volume;
 pub use volume::{VolumesDisplayOptions, VolumesOptions};
 mod volume_mounts;
--- a/internal/engine/tar_stream.rs
+++ b/internal/engine/tar_stream.rs
@@ -1,0 +1,50 @@
+//! The one place a `tar::Builder` is constructed.
+//!
+//! Three paths pack a tar and hand it to Podman: the build context, `cp`, and
+//! the watch sync. All three land in the same server-side reader, so all three
+//! need the same archive dialect, and the way to guarantee that is to have a
+//! single constructor rather than three that agree today.
+//!
+//! `tests/tar_builder_single_site.rs` keeps it single.
+
+/// A `tar::Builder` writing the dialect Podman's archive reader accepts.
+///
+/// `tar::Builder::new` enables sparse-file support by default. With it on, a
+/// file that has holes on disk is written as a GNU sparse entry, whose typeflag
+/// is `b'S'`, decimal 83. Podman refuses that type and the whole transfer dies
+/// on account of one file. Podman's own client packs in Go and never emits the
+/// type, which is why a context `podman build` accepts could be one
+/// `podup build` rejected (#1775).
+///
+/// The refusal has two spellings, measured on Podman 5.7.0, and a search for
+/// either one alone misses half the damage:
+///
+/// - the build endpoint answers `unhandled tar header type 83`, from
+///   `containers/storage/pkg/archive`, and `podup build` exits 1;
+/// - the archive endpoint that `cp` and the watch sync both PUT to answers
+///   `unrecognized Typeflag S`. `cp` exits 1; `watch` only logs
+///   `watch action failed`, keeps running and never delivers the file, so
+///   there the loss is silent.
+///
+/// Nobody puts a sparse file in a build context on purpose. They arrive on
+/// their own, as database files, disk images, virtual machine state, or
+/// anything preallocated with `fallocate`, and `ls` shows nothing unusual, so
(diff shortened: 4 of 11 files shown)


Jaro-c:
Added three integration tests that run against a live Podman, one per affected path, because everything the first commit shipped reads bytes this crate wrote itself. None of it asks the server, which is the only place the failure actually appears.

| Test | What the server says when `.sparse(false)` is reverted |
| --- | --- |
| `build_uploads_a_context_that_holds_a_sparse_file` | `unhandled tar header type 83` |
| `engine_cp_uploads_a_directory_holding_a_sparse_file` | `unrecognized Typeflag S`, and the ordinary file beside the sparse one is lost too |
| `watch_initial_sync_carries_a_sparse_file` | nothing. The sync failure is a warning the loop survives, so the process stays up, keeps saying it is watching and exits 0 while the file never arrives |

I reverted the fix and watched each of the three go red before keeping them.

The watch row is why none of them asserts on an exit code: the assertion is that the bytes are inside the container. Each asserts the arrived length rather than the file's presence, since a transfer that succeeded and truncated the expanded holes would walk past a `test -f`, and each checks the filesystem really left the range unallocated before concluding anything, saying so and stopping when it did not.

They went into two new files rather than into `build_images.rs` and `watch.rs`, which sit at 448 and 296 code lines against a 500 limit with a warning at 300.

Local suite after the addition: 2674 tests, no failures, `fmt` and `clippy --all-features --all-targets` clean.

Issue #1808, opened 2026-09-20
cp verification: a link is confirmed without comparing its target


Jaro-c:
`tree_landed` confirms a symlink entry by the symlink bit in the destination's stat and nothing else, so an existing link to any target satisfies an uploaded link to a different one. The target that was sent is read out of the tar and thrown away, and `PathStat` does not keep libpod's `linkTarget` field either.

The same shape one level down: a regular file is confirmed by its size, so a failed upload over files that already have those lengths is reported as landed. That residual is documented for the single-file path and now applies to a tree.

Found by a review of the 5.9.4 range, by reading. Not reproduced.

## What a fix needs

- Keep the sent link target in `SentEntry`, add `linkTarget` to `PathStat`, and compare them. Podman 5.7.0 sends the field in the stat header (measured 2026-09-18 while building #1790: a link to a file reported `size` as the length of the link text).
- For regular files, decide whether anything beyond the size is worth asking for. libpod's stat has no checksum, so the honest options are a modification time, which the extraction sets from the archive, or nothing. Write down which one and why.

Jaro-c:
Not shipping this in the next release. The work is real but it is not finished, and what it does today is worse than the defect it closes. Leaving the issue open with what two rounds measured, so the next attempt starts from here instead of repeating them.

The branch `feat/pathstat-link-target-1808` carries the work; PR #1823 is closed unmerged.

## What the change does

`PathStat` gains `linkTarget`, `SentKind::Link` carries the target the archive held, and `entry_landed` answers with four states instead of a boolean: a matching target confirms, an absolute target that resolves against the destination directory to the archive's relative target confirms, a target that is answered and differs is refused, and a target the runtime did not answer (absent, or the empty string) falls back to the symlink bit alone.

On Podman 5.7.0 it is right: 2019 unit tests, six live `cp_flags` tests, and three sabotages that each drop exactly the assertions named for them.

## Why it cannot ship

On Podman 6.1.2, `podman-vm (6)` fails with 212 passed and 1 failed, twice, on two different attempts:

```
cp_flags::engine_cp_reports_a_directory_copy_as_landed
a directory copy that landed (the bytes above prove it) must report Ok on
Podman 6 as on Podman 5; got Err(Copy("the upload to /tmp could not be
confirmed: ..."))
```

[...]

PR #1823, opened 2026-09-20
fix(cp): confirm a copied symlink by its target, not by the symlink bit

Jaro-c:
Closes #1808.

## The defect

`tree_landed` confirmed a symlink entry by the symlink bit alone, so an existing link pointing anywhere satisfied an uploaded link. The target read out of the tar was discarded, and `PathStat` did not keep libpod's `linkTarget` either.

## What changed

`PathStat` carries `linkTarget`. `SentKind::Link` carries the target the archive held, read from the tar's linkname header. A link entry is confirmed only when the destination is a symlink **and** its target equals what was sent; a stat that carries no target leaves the entry unconfirmed, which fails closed. A link target that is not valid UTF-8 makes the tree answer "not landed" rather than comparing against a lossy rewrite, the rule the paths already follow.

## What it still cannot see, written down

libpod's stat has no checksum. A failed upload over regular files of the same length is still reported as landed, and a pre-existing link already pointing at the same target passes too. The link path narrows the false positive; it does not close it. That is now in the module doc rather than implied.

## Controls

Both sabotages drop exactly the assertion named for the behaviour:

| Sabotage | Assertion that fell |
|---|---|
| confirm on the symlink bit alone | `a_link_is_landed_only_when_its_target_matches_what_was_sent` |
| accept a stat that carries no target | the same |

The live test also tightened: the symlink check went from `contains("a.txt")`, which a target of `/x/a.txt.bak` would have satisfied, to `assert_eq!` on the trimmed `readlink` output, read one link per exec so each target stands on its own line.

Gates, all rc=0: `cargo test --locked --lib --all-features` (2015 passed), the live `cp_flags::` with `PODUP_REQUIRE_PODMAN=1` (6 passed, so no test skipped itself), `cargo fmt --all --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings`.


Jaro-c:
Closing unmerged. `podman-vm (6)` fails twice with the copy landing and podup reporting it failed, and the reason is not observable because the diagnostic is a `tracing::debug!` the test harness filters out. The branch keeps the work; the state and the next step are written up on #1808.

Issue #1825, opened 2026-09-20
No live test creates a pod with userns options, so the pod path is untested


Jaro-c:
Split out of #1811, which I closed as not reproduced. The behaviour is correct; nothing under test proves it stays correct.

## The gap

`internal/engine/container/userns_tests.rs` builds its containers through a helper that calls `create_and_start` directly. It never creates a pod. So every assertion in that file passes whatever the pod path does, and a regression that stopped forwarding the ID-mapping options into the pod spec would not fail anything.

A pod is a different path from a lone container: `internal/engine/pod/mod.rs` sends the pod spec, `PodSpecGenerator` carries no ID-mappings field of its own, and a member container suppresses its own namespace when it joins the pod. That is the path with no test on it.

## What the test must assert

These are measured on `develop` today, against Podman 5.7.0, with `x-podman-pod: true` and one alpine service. They are the rows the test should pin:

| `userns_mode` | `uid_map` of the member | `id -u` |
|---|---|---|
| `auto` | `0 1 1024` | 0 |
| `auto:size=2048` | `0 1 2048` | 0 |
| `keep-id:uid=321,gid=654` | `0 1 321` | 321 |

Three different options must give three different answers. A test that only asserts `auto` cannot tell a working option path from one that ignores the option and falls back to the default, because the default IS the `auto` answer.

The test must also assert the container is in the pod, by reading its `Pod` field. Without that, a regression that quietly created the container outside the pod would leave every mapping assertion passing while the pod path went unexercised.

## Where it goes

`tests/engine_integration/`, beside the other live tests, guarded the way they are so it runs under `PODUP_REQUIRE_PODMAN=1` rather than skipping into a green.

Jaro-c:
Merged in #1828 into develop. The live test creates a pod with each userns option and reads the member's ID map; the fragmented-allocation sum is pinned by a plain unit test over the two literal maps. Ships with the next release.

Issue #1811, opened 2026-09-20
userns_mode with options fails inside a pod, and the live test bypasses the pod path


Jaro-c:
#1802 made `userns_mode: auto` work for a container by sending the spec's ID-mapping options. A pod is a different path: `internal/engine/pod/mod.rs` sends only `userns` in the pod spec, `PodSpecGenerator` has no ID-mappings field, and a member container suppresses its own namespace when it joins a pod.

So with `x-podman-pod: true` and `userns_mode: auto` (or `auto:size=N`), whether the pod gets its own ID range is unverified, and the same silent no-op #1797 was about may still be there.

The test `internal/engine/container/userns_tests.rs` does not cover it: its helper calls `create_and_start` directly and never creates a pod, so it passes whatever the pod path does.

Found by a review of the 5.9.4 range, by reading. Not reproduced against a daemon.

## What a fix needs

- Measure first: create a pod with `--userns=auto` through the API and read `/proc/self/uid_map` from a member, then the same through podup with `x-podman-pod: true`.
- If the pod spec cannot carry the mappings, say so in `docs/docker-migration.md` beside the `auto` section, and make `podup audit` or the pod path refuse the combination rather than accept it silently.
- A live test that creates a pod, not one that bypasses it.

Jaro-c:
Measured on 2026-09-19 against Podman 5.7.0 rootless, with podup built from `develop` at 5.9.4:

```
podman pod create --userns=auto            -> member reads uid_map: 0 1 1024
podup, x-podman-pod: true, userns_mode auto -> t-1 reads   uid_map: 0 1 1024
```

So the plain `auto` case does reach the pod, and the ID-mapping options #1802 added are carried on the pod spec. That half of this issue is not a defect; I was wrong to assume it from reading.

What is a defect, same measurement:

```
userns_mode: "auto:size=2048"          -> podman API error (HTTP 500): failed to make pod: invalid namespace type auto:size=2048 specified
userns_mode: "keep-id:uid=321,gid=654" -> podman API error (HTTP 500): failed to make pod: invalid namespace type keep-id:uid=321,gid=654 specified
```

The pod path still sends the whole string as the namespace mode, so #1798's split never reached it. Outside a pod both of those start.

So this issue is now two things, both narrower than what I wrote:

1. The pod spec must carry the mode and its options apart, the way the container spec does since #1802.
[...]

Issue #1797, opened 2026-09-19
userns_mode: auto is accepted and has no effect, while the docs and podup audit recommend it


Jaro-c:
`userns_mode: auto` is accepted by podup and has no effect. The container runs in the same user namespace as every other rootless container, which is the opposite of what our own docs and `podup audit` tell people it does.

## Measured on 2026-09-19, Podman 5.7.0 rootless, podup from `/usr/bin`

Podman directly:

```
$ podman run --rm --userns=auto alpine cat /proc/self/uid_map
         0          1       1024
```

A private range, as documented.

podup, one service with `userns_mode: "auto"` and `command: ["sleep","infinity"]`:

```
$ podman inspect <project>-t-1 --format '{{.HostConfig.UsernsMode}}'
                                   <- empty
$ podman exec <project>-t-1 cat /proc/self/uid_map
         0       1000          1
         1     100000      65536
```

That is the default mapping, byte for byte the same as a container with no `userns_mode` at all. No error, no warning.

## Why it matters

- `docs/docker-migration.md`, section "`userns_mode: auto` gives each container its own UID range", recommends exactly this to anyone who wants containers of one project isolated from each other on shared volumes and after an escape. It does not happen.
- `podup audit`'s `no_userns` check (`internal/audit/checks.rs`) flags services without `userns_mode` and points at that section. Its doc comment and message also say `auto` is "the default behaviour when the field is absent". It is not: with the field absent Podman used the rootless default above (`UsernsMode` empty, no `userns` line in either `containers.conf` on this host).

So an operator who follows the audit ends up with a green audit and no isolation.

## What the fix needs

- Find what the SpecGenerator needs for `auto` over the REST API. `podman run --userns=auto` works, so the CLI sets something podup does not; I have not looked yet whether it is `idmappings`, a separate field, or the `value` half of the namespace.
- A live test in the engine integration suite that reads `/proc/self/uid_map` from a container with `userns_mode: auto` and asserts the first line is not `0 <host uid> 1`.
- Correct the `no_userns` doc comment and message so they state the real default.
- Related and filed separately: `userns_mode` values with options, such as `auto:size=65536` that the same doc section recommends, fail with HTTP 500.

Jaro-c:
Landed on `develop` in #1802. `userns_mode: auto` now carries the ID-mapping options, measured against Podman 5.7.0 with a control container: without them the two `auto` tests fail.

PR #1828, opened 2026-09-20
test(engine): a live test creates a pod and reads the member's ID map

Jaro-c:
Closes #1825, split out of #1811 which I closed as not reproduced.

## The gap

`internal/engine/container/userns_tests.rs` builds its containers through a helper that calls `create_and_start` directly and never creates a pod, so every assertion in it passes whatever the pod path does. A pod is a different path: `internal/engine/pod/mod.rs` sends the pod spec, `PodSpecGenerator` carries no ID-mappings field of its own, and a member container suppresses its own namespace when it joins the pod. Nothing covered it.

## What the test asserts

A project with `x-podman-pod: true`, reading `/proc/self/uid_map` and `id -u` from the member:

| `userns_mode` | `uid_map` | `id -u` |
|---|---|---|
| `auto` | `0 1 1024` | 0 |
| `auto:size=2048` | `0 1 2048` | 0 |
| `keep-id:uid=321,gid=654` | `0 1 321` | 321 |

**Three rows rather than one, and that is the point.** A path that dropped the option and fell back to the default would answer `0 1 1024`, which is exactly what `auto` answers. A test that asserted only `auto` could not tell a working option path from a silent no-op. Only the other two rows separate them.

It also asserts the container's `Pod` field equals the project's pod, so a regression that quietly created the container outside the pod cannot leave every mapping assertion passing while the pod path goes unexercised.

## The control

I sabotaged the production path, not the test: made `pod_userns` keep only the base mode and discard everything after the colon, which is the silent no-op #1797 was about.

    userns-pod-auto-size-2048: uid_map size must be 2048 for userns_mode
    "auto:size=2048", got "         0          1       1024"
      left: 1024
     right: 2048

The pod fell back to the `auto` default and the test named the case, the mode, the expected size and the one it got. That is the regression this test exists to catch, and it catches it.

## Tear-down

A guard that runs on drop, so a panicked assertion cannot leave pods holding subordinate ID ranges. A host whose range is exhausted fails every later run with `not enough unused IDs in user namespace`, which reads like a podup defect and is not: that is what an exhausted range looked like on this machine this morning, and it cost an hour before the cause was found.

Gates, all rc=0: the live test with `PODUP_REQUIRE_PODMAN=1` (1 passed, 31.74s, so it ran rather than skipped), `cargo test --locked --lib --all-features` (2026 passed), `cargo fmt --all --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings`.


Jaro-c:
Not in 5.9.5. Leaving this open with what is done and what is not, so the next session does not have to reconstruct it.

## Fixed since the last push

The test asserted on the first line of `/proc/self/uid_map`. It passed twice on this host and then failed once with no code change between:

    userns-pod-auto-size-2048: uid_map size must be 2048 for userns_mode
    "auto:size=2048", got "         0          1       1024
          1024       2049       1024
    "

podup had delivered the 2048 IDs, in two ranges of 1024, because no contiguous block was free at that moment. The `auto` family now sums the size across every range. `keep-id` keeps the first-line reading: its map is three lines where the first is the requested range and the other two are plumbing, so the sum there would expect 65537 instead of 321.

Four live runs in a row from the author, two more from me, all green.

## What still blocks it

**CodeQL, and inline suppression does not work.** Four `rust/cleartext-logging` alerts fire on the assertion messages because the interpolated values have uid-style names. The value is the ID map of an alpine container this test creates and destroys in the same function, the only sink is the assertion message, and the message is what makes the control worth having.

I added `// codeql[rust/cleartext-logging]` comments and pushed. The check stayed red on the new commit, with the same four alerts at their new line numbers. So the comments record the reasoning for a reader but do not clear the check.

The options left are a `codeql-config.yml` query filter scoped to `rust/cleartext-logging` under `tests/**`, or dismissing the four alerts in code scanning. That is a policy decision about how security findings in test code are handled in this repository, and it should be its own pull request with its own review rather than something decided to unblock a test on a release day.

## What is NOT proven, and should be first next time

`total_size` has no unit test. It was verified with a standalone harness that is not in the repository, so nothing committed holds it.

I sabotaged the assertion back to reading the first line and the live test still passed, because the host currently hands out a contiguous 2048 block. The sabotage was inert, which means this fix is not demonstrated by the suite as it stands: the failure it repairs only appears when the subuid range is fragmented, and that is not reproducible on demand.

A unit test over the two map strings, the contiguous `0 1 2048` and the split `0 1 1024` / `1024 2049 1024`, would make the control deterministic and make that sabotage bite every time. That is the first thing to add.

## What is proven

The test catches the regression it exists for. Sabotaging `pod_userns` to keep only the base mode and discard the options makes it fail with the case, the mode, the expected size and the one it got.

Issue #1795, opened 2026-09-19
cp of a source ending in /. copies the directory itself, not its contents


Jaro-c:
`podup cp payload/. SERVICE:/tmp` should copy the contents of `payload` into `/tmp`, the way `docker cp` and `podman cp` treat a source ending in `/.`. It copies the directory itself instead, and the files land in `/tmp/payload/`.

## Why

`pack_path` in `internal/engine/copy/archive.rs` names the archive's top entry with `src.file_name()`. Rust's `Path` drops a trailing `.` component, so `Path::new("payload/.").file_name()` is `Some("payload")`, and the tree is packed under `payload/`. Nothing on the way from the CLI to `pack_path` handles the `/.` suffix: I searched for it and found no match.

I found this by reading, during a review of #1790, and have not run it against a container yet. It is not introduced by #1790; it is on `develop` today.

## What the fix needs

- Detect the trailing `/.` on the source as written, before it becomes a `Path`, since the `Path` has already lost it.
- Pack the directory's children at the top of the archive rather than under the directory's name.
- A live test in `tests/engine_integration/cp_flags.rs` that copies `payload/.` and asserts the files are at `/tmp/a.txt`, not `/tmp/payload/a.txt`, plus the same for the host direction if `cp SERVICE:/path/. host` has the same shape.

Jaro-c:
Fixed in #1822, merged to `develop` as `ea33e38`. GitHub does not auto-close from a non-default branch, so closing by hand.

Jaro-c:
Correction: the closing note on this issue and the commit message both say the container-to-host direction carries the same divergence. Measured on the published 5.9.5 binary against `podman cp`, in four cases (destination existing or not, source with and without the trailing `/.`), podup and podman agree in all four. That direction was never wrong. Details and the table are on #1830.

Issue #1777, opened 2026-09-10
cp of a directory reports failure against Podman 6 even when the copy lands


Jaro-c:
## What happens

`cp` of a directory into a container reports failure against Podman 6, whatever it actually did on disk:

```
cp error: the upload to /tmp could not be confirmed: the container runtime closed the connection
without a response and the destination did not change. The copy may or may not have landed;
check /tmp in the container.
```

Measured on the `podman-vm (6)` lane, Podman 6.1.1, in the run for #1776. The same copy against Podman 5.7.0 returns `Ok`.

## Why

Podman 6 answers an archive PUT by applying it and closing the connection without a response, which arrives as an `IncompleteMessage`. `internal/engine/copy.rs:440` already knows this and recovers by re-verifying the destination instead of trusting the transport. The recovery needs something to compare against, and that comes from `uploaded_entry_size` (`internal/engine/copy.rs:482`):

```rust
pub(super) fn uploaded_entry_size(src: &std::path::Path) -> Option<u64> {
	std::fs::metadata(src)
		.ok()
		.filter(std::fs::Metadata::is_file)
		.map(|m| m.len())
}
```

For a directory that filter yields `None`, so `expected` is `None`, `landed` is false, and the copy fails closed. Its own doc comment says as much and calls it deliberate: a directory entry's own size says nothing about whether its children arrived.

Failing closed is the right instinct. Reporting a copy as failed when it landed is still a defect, because the caller cannot tell that from a copy that did not, and the message says so out loud.

## Why nothing caught it

Every `cp` test in the suite copied a single file. Nothing copied a directory against Podman 6 until I added a directory case in #1776 for an unrelated reason, and that is what turned it up. The lane was green because the case did not exist.

## What a fix has to answer

A directory has no single entry to compare, so the confirmation needs a different question. The candidates I see:

- Ask about one known child rather than the directory: pick the first regular file in the packed tree and check its size at the destination. Cheap, and it catches a truncated or refused stream, but it says nothing about the other children.
- Ask about the whole tree: walk the archive we just sent and stat every regular file in it. Exact, and the cost is one stat per file, which is a lot of round trips on a large tree.
- Ask Podman for a listing of the destination once and compare names and sizes against what we packed. One round trip, and it needs the endpoint to give a recursive listing, which is worth checking before designing around it.

I have not chosen. What is not acceptable is treating an unconfirmable directory copy as success, which is the failure mode this currently avoids.

## Acceptance

[...]

PR #1830, opened 2026-09-20
release: 5.9.5

Jaro-c:
Five commits since 5.9.4, one of them visible to a user.

## What ships

**`cp` of a source whose last component is `.` copies the directory's contents** (#1795). `podup cp payload/. SERVICE:/tmp` put the files in `/tmp/payload/` before; they land in `/tmp` now, the way `docker cp` and `podman cp` treat it.

Measured against podman into a destination directory that already exists, because a non-existent destination flattens on its own and would have confounded the reading:

| Source | podman | podup before | podup now |
|---|---|---|---|
| `payload` | `dest/payload/a.txt` | same | same |
| `payload/.` | `dest/a.txt` | `dest/payload/a.txt` | matches |
| `payload/./` | `dest/a.txt` | `dest/payload/a.txt` | matches |
| `.` from inside | `dest/a.txt` | `dest/payload/a.txt` | matches |

The cue is read with `std::path::is_separator`, so the backslash shapes fire on Windows and not on Unix, where a backslash is an ordinary filename character. A source that is not a directory is refused the way podman refuses it. The container to host direction is unchanged and still lands the contents under the source's basename.

## What does not ship but mattered

Four CI fixes with nothing user-visible in them. Two are worth naming:

`branch health` read the newest completed run on a branch and could answer for a commit nobody tested; it now reads the run for the branch's head (#1824). That change then broke both callers, because it handed the pushed commit to whichever branch was being asked about, and the two callers are cross-branch by design (#1827). Together they took `branch health (main)` on a push to `develop` from exhausting 32 attempts and failing after 8m18s down to **7 seconds**, and closed a loop where a red on one branch held the other red.

## Deliberately left out

- **#1808**, the symlink target confirmation. Right on Podman 5.7.0, and on Podman 6.1.2 it reports a landed copy as failed, which is the #1777 shape closed two days ago. PR #1823 closed unmerged; the issue carries the state and the next step.
- **#1825**, the live pod userns test. Its control is proven, but the fix for its own defect is not demonstrated by the suite: the failure it repairs only appears when the host's subuid range is fragmented, and that is not reproducible on demand. PR #1828 stays open with that written down.


Jaro-c:
Correction to this pull request's description, and to the commit message of #1822 which is now in `main`.

Both say the container-to-host direction still lands the contents under the source's basename, or "has the same divergence". **That is false, and I did not measure it before writing it.** I took it from the working notes of the change rather than from the daemon.

Measured on the published 5.9.5 linux-x86_64 binary against Podman 5.7.0, four cases, comparing podup with `podman cp` on the same source tree:

| Case | podup | podman cp |
|---|---|---|
| `a:/orig/. dst` where `dst` exists | contents in `dst` | contents in `dst` |
| `a:/orig dst` where `dst` exists | wrapped as `dst/orig` | same |
| `a:/orig/. dst` where `dst` does not exist | contents in `dst` | contents in `dst` |
| `a:/orig dst` where `dst` does not exist | contents in `dst` | contents in `dst` |

They agree in all four. The container-to-host direction handles the trailing `/.` correctly and always did; there was nothing to close there.

The 5.9.5 changelog entry is not affected: it says "The container to host direction is unchanged", which is accurate, since the change did not touch it.

What this costs: #1822 shipped with a "what this does not close" paragraph describing a defect that does not exist, and that paragraph is in `main`'s history where it cannot be edited. Anyone reading that commit later will go looking for a divergence that is not there. This comment is the correction of record.

Issue #1843, opened 2026-09-20
The integration suite leaks one tagged image per run, and prune cannot reclaim them


Jaro-c:
**Component:** tests

## What is on this machine

```
$ podman images --format '{{.Repository}}' | grep -cE '^docker.io/library/t[0-9]+-'
186
$ podman images -f dangling=true --format '{{.Repository}}' | grep -cE '^docker.io/library/t[0-9]+-'
0
```

186 images named after the integration suite's per-run prefix (`t<pid>-clbld-web`, `t<pid>-...`), and **none of them is dangling**, so `podman image prune` never removes them. Only `prune -a` would, and that also removes every legitimate tagged image on the host, which is not something to run by reflex.

I am not putting a disk figure on it. `podman images` reports each image's total including shared base layers, so summing the column would produce a number that is not the space they occupy.

## Why this is worth an issue rather than a cleanup

The same leak, in its container form, cost an hour of this session and produced a false bug report. On 2026-09-20, `podman pod create --userns=auto` failed with

    not enough unused IDs in user namespace

on every form, including plain `auto`. That was read as podup rejecting the userns options and drove the work on #1811. It was not podup: 231 abandoned build containers and 823 orphan volumes, left by previous test runs, were holding the host's 65536 subordinate IDs. #1811 was closed as not reproduced once the range was freed.

So the leak is not only untidy. It changes what the test suite measures on a machine that has run it many times, and it produced a wrong diagnosis that took a measurement to undo.

## What to decide

Each run tags an image with its own pid, so nothing ever reuses or overwrites them: the set grows without bound. The options are the obvious ones, and the choice is a project decision rather than a defect to patch blindly:

- remove the image in the same guard that removes the run's containers, the way #1825's live test does for its pods;
- or keep tagging per run but give the suite a documented cleanup entry point, so a developer has one command rather than a hand-written `grep` over `podman images`.

Whichever is chosen, the containers and volumes deserve the same treatment: the pods in #1828's test are already cleaned on drop, and that pattern is the one that worked.

Jaro-c:
## How this stays fixed

This is itself a test-hygiene issue, so the control belongs in the harness rather than in a new test.

The assertion that would have caught it: after the integration suite finishes, no image, container, volume or network carrying the suite's per-run prefix remains. That is one check at the end of a run, and it fails loudly on the machine that leaked rather than months later on an unrelated bug hunt.

[...]

PR #1860, opened 2026-09-21
test: scan for leaked fixtures after the integration suite exits

Jaro-c:
The integration suite left resources on the host on every run and nothing noticed: 125 containers, 189 networks and several hundred images had piled up on one machine, and the containers alone held 58 of the 128 inotify instances a user may open (#1843).

**Why a separate binary.** The first version of this PR scanned from inside `engine_integration` and failed on `podman-vm (5)`: cargo runs one binary's tests in parallel, so a `z_`-named scanner reported two resources of a test that finished after it. The scan now lives in `integration_leak_scan`, run as a second `cargo test` process once the suite has exited, whether it passed or not. `engine_integration` hands over its PID through `target/podup-leak-scan-pid`.

**The lane fails** when the scan reports a leak, when its marker is missing, and when the PID file was never written (`rc=nopid`), since the scan exits cleanly without a PID and would otherwise pass having scanned nothing. I drove the verify snippet with all four marker shapes before pushing.

**Guarding the scanner:** one test plants a network under its own prefix and requires the scanner to name it; another drives `podman` into a non-zero exit and requires that to fail the scan rather than read as an empty list. Emptying the scanner or making it swallow the exit each fails one of them.

**What it found, and what is fixed here:** seven leaks on its first local run. Two were a podup defect, fixed in #1868. Five were tests that never tore down; they now bind a `DownGuard` that runs `down -v` on drop and prints the error if `down` fails. Two build tests hold a `TestImage` guard.

**Measured:** the whole live suite, 218/218 on Podman 5.7.0, then the scan with that run's PID: nothing. A manual listing of containers, networks, volumes, pods and images with the run's prefix was empty.

Closes #1843


Jaro-c:
`podman-vm (5)` is red and the failing test is the one this PR adds: `z_no_leaks_after_suite`, 214 passed / 1 failed. It is not the suite leaking — it is the detector measuring while the suite is still running.

```
test z_no_leaks_after_suite ... FAILED
test x_podman_pod::up_with_pod_creates_one_pod_with_every_service ... ok
```

That ordering is from the run's own log: a test finished **after** the detector had already panicked. And two of the twelve resources it named belong to exactly that test:

```
the suite left resources with the per-run prefix t3921-:
  image: docker.io/library/t3921-wrb-web
  image: docker.io/library/t3921-clbld-web
  container: t3921-pod-one-web-1     <- x_podman_pod, still running
  container: t3921-pod-one-db-1      <- x_podman_pod, still running
  volume: t3921-idv_data
  network: t3921-clrm_default
  ... six more networks
```

The `z_` prefix only orders the *name*, and cargo runs integration tests in parallel, so it does not order the *execution*. "After the suite" is not a moment that exists inside the suite when the runner is parallel: the last-named test still starts alongside the others.

So the assertion cannot live where it is. Two shapes that do work, and the choice is worth making deliberately rather than patching the symptom:

1. **Move the scan out of the test binary.** The suite writes its per-run prefix to a file, and a CI step after `cargo test --test engine_integration` scans for that prefix. This is the honest version of "after the suite", because the process has exited. It costs a CI step and the prefix hand-off.
2. **Keep it in-process behind a barrier** — an in-flight counter every test increments through a guard, with the detector waiting for zero. This is more code in the test harness and has its own race: the detector can read zero before the other tests have started.

I have not applied either. Also unresolved from the original review: nothing guards the scanner itself — replacing `find_per_run_leaks` with an empty body left the detector passing, so whichever shape this lands in needs a test that the scanner detects a planted resource.

One thing this run does establish: the images `t3921-wrb-web` and `t3921-clbld-web` were present at scan time. Whether they are the real leak from #1843 or just tests still in flight is not something this log settles — the captured output only names 29 of the 215 tests, so I cannot say when their tests finished.

Issue #1835, opened 2026-09-20
audit has no check for a port published without a host IP


Jaro-c:
## What problem does this solve?

`podup audit` says nothing about a port published without a host IP. Under rootless Podman, `- "5432:5432"` binds every interface, so a database that the author believed was local is reachable from whatever network the machine is on. On a laptop that moves between networks, that is the whole stack exposed to the room.

The audit already covers the runtime side of hardening thoroughly. This is the one exposure that happens before a single capability matters, and it is visible in the compose file without touching Podman.

## What I would expect

A check that fires on a `ports:` entry with no host IP, and on one that spells `0.0.0.0` explicitly. `127.0.0.1:5432:5432` and `[::1]:5432:5432` stay quiet.

`Service::ports` is already a `Vec<PortMapping>` in `internal/compose/types/service.rs`, so the short, long and flow forms are parsed before the check would see them.

## Note on false positives

A service that is *meant* to be public — an ingress, a reverse proxy — would fire this every time. Two ways out, either is fine by me: keep it as a finding and let the project silence it the way any other finding is handled, or scope the check to services that also publish a well-known data-engine port. I would rather have the noisy version than not have the check: it is the one people get wrong on the first deploy and never revisit.

**podup version:** v5.9.4
**Podman:** 5.7.0
**OS:** Ubuntu 26.04.1 LTS

Jaro-c:
Merged to `develop` in #1861. `podup audit` now raises `port_published_on_all_interfaces` for a port published without a host IP, reading the same predicate the parse-time warning on `up`/`config` uses, so the two surfaces cannot disagree. The check has a `CHECK_REGISTRY` entry.

PR #1861, opened 2026-09-21
feat(audit): flag a port published on every host interface

Jaro-c:
`podup audit` had no opinion on port exposure, while `up` and `config` have warned about it since the port-exposure diagnostic landed. An operator running the audit as a deployment gate got a pass on a compose file that publishes Postgres on every interface.

The check reads the same predicate the diagnostic does rather than restating it: `ports_published_on_all_interfaces` is now the single source of truth in `internal/compose/diagnostics/ignored_fields.rs`, and both surfaces format their own message from the `PortExposure` it returns. A future compose shape that counts as published reaches the audit and the warning together.

Thresholds are the diagnostic's, unchanged. An explicit host IP is not flagged, including `0.0.0.0` and a private LAN address; a container-only short form is not a publish; a range emits one finding with the range string as its label.

One behaviour changes in the diagnostic: `":5432"` used to emit `port  is published on every interface`, with the label missing. It is now skipped, and a test pins both halves.

**Sabotage, measured:** emptying the shared predicate stops the binary emitting the finding (`podup audit` on a `5432:5432` fixture goes from one `port_published_on_all_interfaces` line to zero) and fails six of the sixteen new audit tests plus four of the diagnostic's. Removing the empty-host guard fails the test that pins it.

Note on running them: the audit tests live in the binary target, so `cargo test --lib` does not include them. `cargo test` with no target flag does.

Closes #1835


Jaro-c:
Rebased onto `develop` after #1859 landed the check registry. The check function moved into `check_fns.rs` and has a `CHECK_REGISTRY` entry; both dirty fixtures the registry tests use now publish `5432:5432`, because those tests require every entry to fire, and without it they failed naming the new id.

One correction to the sabotage numbers in the description. I read them with a plain `cargo test`, which stops at the first target that fails, so the binary target never reported. Re-read with `--no-fail-fast`: emptying the shared predicate fails **fourteen** tests, across the audit, the diagnostic, the registry and the `up`/`config` warning — not ten.

Issue #1834, opened 2026-09-20
audit has no check for a bind mount of a sensitive host path


Jaro-c:
## What problem does this solve?

`podup audit` has eleven checks and none of them looks at `volumes:`. A service can mount a sensitive host path and still come back clean, which undoes what the other ten checks achieve: a container with `read_only: true`, `cap_drop: [ALL]`, `no-new-privileges` and its own user namespace, mounting the host root or the podman socket, is not a hardened container.

```
$ grep -c 'volumes' internal/audit/checks.rs
1        # and that one occurrence is a comment
```

This is the gap I would close first, ahead of anything else on the audit surface, because it is the one that silently cancels the rest.

## What I would expect

A check that fires on a bind mount whose host source is a sensitive path. A starting list, roughly in order of how much damage each one allows:

- the container runtime socket: `/run/podman/podman.sock`, `/run/user/*/podman/podman.sock`, `/var/run/docker.sock` — a mount that hands over control of the engine
- `/` and `/etc`
- `/proc`, `/sys`, `/dev`
- `/boot`
- the invoking user's home directory

`:ro` reduces some of these but not the socket, so read-only should probably not suppress the finding by itself.

The data is already parsed: `Service::volumes` is a `Vec<VolumeMount>` in `internal/compose/types/service.rs`. Named volumes and relative project paths should not fire — only absolute host paths on the list.

## Why now

I lean on `podup audit --strict` as the gate that decides whether a deployment is hardened. Today that gate says "no findings" for a compose that mounts the socket, so the answer it gives is not the answer I need. Everything else I asked of the audit it does well; this is the hole under it.

**podup version:** v5.9.4
**Podman:** 5.7.0
**OS:** Ubuntu 26.04.1 LTS

Jaro-c:
Merged to `develop` in #1865. `podup audit` now raises `sensitive_bind_mount` for a bind of a container runtime socket (with stronger wording) or of `/proc`, `/sys`, `/dev`, `/etc`, `/boot`, `/root` and the runtime directories that hold a socket. The rootless Podman socket, `/run/user/<uid>/podman/podman.sock`, is matched by shape; nothing else under `/run/user/` is flagged. `:ro` still fires.

PR #1865, opened 2026-09-21
feat(audit): flag a bind mount of a sensitive host path

Jaro-c:
`podup audit` had no opinion on bind mounts. A compose file that mounts the container runtime socket into a service — which hands it the ability to start sibling containers as the daemon's user — passed `--strict` silently.

`sensitive_bind_mount` fires on:
- **the runtime socket**, matched exactly, with stronger wording (the file is the attack);
- **a short list of host directories**, matched as a prefix so a subpath is the same finding: `/proc`, `/sys`, `/dev`, `/etc`, `/boot`, `/root`, and the runtime directories that hold a socket.

`:ro` still fires. Relative paths and the project directory do not.

**The rootless socket** is matched by shape: `/run/user/<uid>/podman/podman.sock` is the one a rootless operator has and the one podup connects to by default, and the uid varies. The first version missed it while its comment called `/run/podman/podman.sock` the rootless socket — that is the system socket, and the comment now says so. The uid must be all digits; nothing else under `/run/user/` is flagged (session bus, audio, display sockets are mounted on purpose by desktop containers).

| Mount | Findings |
|---|---|
| `/run/podman/podman.sock` | 1 |
| `/var/run/docker.sock` | 1 |
| `/run/user/1000/podman/podman.sock` | 1 |
| `/run/user/1000/pulse/native` | 0 |

Measured on the built binary.

Built on top of #1859's registry: the check lives in its own `check_sensitive_bind.rs` and has a `CHECK_REGISTRY` entry. Both dirty fixtures now plant a rootless socket, because #1859's test requires every registry entry to fire. It did its job here: without the fixture line the test failed and named the new id.

**Sabotage:** rootless matcher returning nothing fails both rootless tests and the registry test; dropping the all-digits uid rule fails the test that pins what is left alone.

Closes #1834


Jaro-c:
Rebased onto `develop` after #1861 landed. Both add a `CHECK_REGISTRY` entry and a line to the two dirty fixtures, so they collided; the fixtures now carry both the port and the rootless socket.

Sabotage re-read on the rebased tree with `--no-fail-fast` (plain `cargo test` stops at the first failing target): making the rootless matcher answer nothing fails **four** tests — both rootless tests, plus the registry's unit test and its integration twin — not the three the description says. Full suite green at 2707.

Issue #1874, opened 2026-09-22
The live watch tests fail under heavy test parallelism, on develop too


Jaro-c:
**Component:** engine / watch (live integration tests)

## What happens

The live `watch` group in `tests/engine_integration` passes at low parallelism and loses two or three tests at high parallelism. It does this on `develop`, not only on a branch. Measured while reviewing #1873, same machine, back to back:

| tree | `--test-threads` | result | wall |
|---|---:|---:|---:|
| `develop` | 1 | 10/10 | 124 s |
| `develop` | 4 | 10/10 | 33 s |
| `develop` | 16 | 7/10 | 132 s |
| #1873 | 1 | 10/10 | 119 s |
| #1873 | 4 | 10/10 | 35 s |
| #1873 | 8 | 7/10 | 133 s |

A run at the default thread count on `develop` also came back 8/10 once.

The tests that drop out vary between runs:

- `watch_delete::watch_sync_propagates_host_deletions_to_the_container`: "the host-side deletion did not propagate"
- `watch_tests::watch_rebuild_recreates_the_container_from_the_new_image`: "the container still serves the old image"
- `watch_tests::watch_sync_and_restart_does_both`: "sync+restart did not copy the file"

Each passes when run on its own.

When the tests fail, the whole group also takes about four times longer, which points at something waiting rather than something wrong.

## Why it is worth fixing

Two reasons, one about the tests and one possibly about podup:

1. A flaky group trains everyone to rerun it, and a real regression in `watch` would hide in the reruns. Diagnosing #1873 cost an evening because of it.
2. If the cause is in podup and not in the test, `podup watch` on a loaded machine may be missing file events, which is a user-visible defect.

Separately: the same tests fail, with no error in the log, when the host has exhausted its inotify instances (measured when 125 leaked test containers held most of them; see #1843). If `podup watch` cannot create its watcher and does not say so, that is a defect of its own, whatever explains the parallelism case.

## Not yet established

A hypothesis from the investigation, **not verified**: the watcher's setup or its event receive runs on the async runtime's worker threads, so under load the watch task is scheduled late and a file change lands before the watch is registered. Nothing measured shows an event actually being dropped. The first step is to instrument the watcher setup and event delivery under load and find out, before changing the watch loop.

Jaro-c:
[...]

PR #1887, opened 2026-09-22
test(watch): fail with the watcher's own error when it dies early

Jaro-c:
The live watch tests failed under heavy parallelism with messages that blamed the sync or the rebuild (#1874). Instrumenting one file change end to end showed the event was neither lost nor late: **the watcher was never created**. `RecommendedWatcher::new` returned `Too many open files (os error 24)` because the user's inotify instances were exhausted (128 by default, 109 in use when measured), which happens when eight watchers start at once on a machine near its limit.

**podup itself is correct:** `podup watch` prints `podup: error: watch error: Too many open files` and exits non-zero. **The tests were silent:** they spawned `engine.watch()`, never checked the task, polled the container until their deadline, and failed on the condition they were waiting for.

`poll_with_watch` now owns that poll and checks on every tick whether the watch task finished; if it did, the test fails with the task's own result (`watch() returned early: <error>`). Every watch test that spawns `watch()` uses it; no container assertion changed.

This does not make the tests pass without free inotify instances; it makes them say why they fail.

**Proof:** a self-test with a task that fails at once requires the helper to report it; making the helper ignore a finished task fails it. The live watch group: 11/11 twice at four threads. Full suite: 2780.

**Not covered by a test:** the CLI's print-and-exit when the watcher cannot be created. Reproducing it means exhausting the user's inotify instances, which every process of that user shares, so it was verified by hand during the investigation and not pinned.

Closes #1874


Jaro-c:
Confirmation after raising the machine's limit, which is what the diagnosis predicted:

```
$ sudo sysctl fs.inotify.max_user_instances=512   # was 128, 109 in use
$ cargo test --features test-helpers --test engine_integration watch -- --test-threads=8
test result: ok. 11 passed; 0 failed; finished in 23.26s
$ ... --test-threads=16
test result: ok. 11 passed; 0 failed; finished in 23.28s
```

Same tests, same tree, same Podman: 11/11 at the thread counts that failed 7/10 before, and 23 s instead of 133 s. The failures were inotify exhaustion, not a race in the watch loop. This PR does not fix that (the limit is the machine's); it makes the tests name it when it happens.

